Issue #21, opened 2019-04-04
Field-handler should receive field-diffs, not the object-diffs


nolar:
## Expected Behavior

The field-handler (`@kopf.on.field`) receives the `diff` object with the diffs specific for that field, same as the `old` & `new` values. And the field paths should be relative to the handled field.

## Actual Behavior

The `diff` object contains the diffs for the whole object (all fields), and relative to the object's root.

Detected while writing the documentation (though was marked with a TODO in the source code).

PR #22, opened 2019-04-04
Pass the field diff to the field handlers, not the obj diff

nolar:
The diff field for the `@kopf.on.field` handler contains the whole obj diff, but should contain the field diff: i.e. only the records related to the handled field, with the field names relative to the field's root, not to the object's root.

> Issue : #21

### DIFF
--- a/kopf/structs/diffs.py
+++ b/kopf/structs/diffs.py
@@ -19,6 +19,16 @@ def resolve(d: Mapping, path: DiffPath):
     return result
 
 
+def reduce_iter(d: Diff, path: DiffPath) -> Generator[DiffItem, None, None]:
+    for op, field, old, new in d:
+        if field[:len(path)] == path:
+            yield (op, field[len(path):], old, new)
+
+
+def reduce(d: Diff, path: DiffPath) -> Diff:
+    return type(d)(reduce_iter(d, path))
+
+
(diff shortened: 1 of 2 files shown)


kopf/structs/diffs.py
```diff
@@ -19,6 +19,16 @@ def resolve(d: Mapping, path: DiffPath):
...
 
 
+def reduce_iter(d: Diff, path: DiffPath) -> Generator[DiffItem, None, None]:
+    for op, field, old, new in d:
+        if field[:len(path)] == path:
+            yield (op, field[len(path):], old, new)
+
+
+def reduce(d: Diff, path: DiffPath) -> Diff:
+    return type(d)(reduce_iter(d, path))
```
samurang87: so Diff is callable?

nolar: `Diff` is not actually a type, it is a declaration of a type with the new Python3 type hinting and annotations (see [here](https://docs.python.org/3/library/typing.html)) — i.e. an "interface" in OOP terms.

The real types matching `Diff` are `list`, `tuple`, and other [Sequences](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes). They are indeed callable (as any classes); and when called, they create a new instance from the provided argument(s).

nolar: This pattern — `type(oldval)(newval)` — is basically about creating an instance of the same type as was fed into the function, without many if's.

Issue #25, opened 2019-04-08
Operator freezes while exiting after an error in the watching/queueing cycle


nolar:
## Expected Behavior

When an error happens in the watching/queueing coroutines, the process exits, and the pod is restarted by Kubernetes (or it just exists if executed locally).

## Actual Behavior

In some cases, the process freezes after the exception, and no new events are handled, nothing is logged.

## Steps to Reproduce the Problem

Uncertain, but:

  1. Simulate an error in the watching cycle, e.g. such as #10

nolar:
The issue is _presumably_ fixed in #27.

"Presumably" means that it was one certain way of freezing the operator forever with no reaction. This way is now fixed. But it is unclear how this way could be triggered and activated: the `kubernetes.watch.Watch().stream()` call never ends normally, since it has `while True` inside.

Which, in turn, means that there could be other reasons and ways of freezing. We need to catch them first, and investigate if it happens again.

PR #27, opened 2019-04-12
Replace built-in StopIteration with custom StopStreaming for API calls

nolar:
Prevent the operator from freezing on the end of the watch-event stream.

> Issue : #25 

Python3 has issues with `StopIteration` exception from inside the async coroutines. This issue is addressed in [PEP-479](https://www.python.org/dev/peps/pep-0479/).

Briefly, this code works and raises a `RuntimeError` as per PEP-479:

```
import asyncio

async def fn(src):
    print(next(src))
    print(next(src))  # raises StopIteration

loop = asyncio.get_event_loop()
coro = fn(iter([100]))
loop.run_until_complete(coro)
```

```
RuntimeError: coroutine raised StopIteration
```

But this code prints an exception and hangs forever (here, we execute synchronous `next` in a thread pool aka asyncio "executor"):

```
import asyncio

async def fn(src):
    print(await loop.run_in_executor(None, next, src))
    print(await loop.run_in_executor(None, next, src))  # raises StopIteration

loop = asyncio.get_event_loop()
coro = fn(iter([100]))
loop.run_until_complete(coro)
```

```
TypeError: StopIteration interacts badly with generators and cannot be raised into a Future
```

After which it hangs forever. Therefore, using this `next`-in-executor technique is not safe and should be avoided.

This PR replaces the `StopIteration` with custom `StopStreaming`, and handles it accordingly.

### DIFF
--- a/kopf/reactor/watching.py
+++ b/kopf/reactor/watching.py
@@ -5,8 +5,17 @@ Kubernetes client's watching streams are synchronous. To make them asynchronous,
 we put them into a `concurrent.futures.ThreadPoolExecutor`,
 and yield from there asynchronously.
 
+However, async/await coroutines misbehave with `StopIteration` exceptions
+as raise by the `next` method: see `PEP-479`_.
+
+As a workaround, we replace `StopIteration` with our custom `StopStreaming`
+inherited from `RuntimeError` (as suggested by `PEP-479`_),
+and re-implement the generators to make them async.
+
 All of this is a workaround for the standard Kubernetes client's limitations.
 They are not needed if the client library is natively asynchronous.
+
+.. _PEP-479: https://www.python.org/dev/peps/pep-0479/
 """
 
 import asyncio
@@ -15,10 +24,30 @@ import logging
 logger = logging.getLogger(__name__)
 
 
+class StopStreaming(RuntimeError):
+    """
+    Raised when the watch-stream generator ends streaming.
+    Replaces `StopIteration`.
+    """
+
+
+def streaming_next(src):
+    """
+    Same as `next`, but replaces the `StopIteration` with `StopStreaming`.
+    """
+    try:
+        return next(src)
+    except StopIteration as e:
+        raise StopStreaming(str(e))
+
+
 async def streaming_aiter(src, loop=None, executor=None):
     """
     Same as `iter`, but asynchronous and stops on `StopStreaming`, not on `StopIteration`.
     """
     loop = loop if loop is not None else asyncio.get_event_loop()
     while True:
-        yield await loop.run_in_executor(executor, next, src)
+        try:
+            yield await loop.run_in_executor(executor, streaming_next, src)
+        except StopStreaming:
+            return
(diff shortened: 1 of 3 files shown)


kopf/reactor/watching.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    Same as `next`, but replaces the `StopIteration` with `StopStreaming`.
+    """
+    try:
+        return next(src)
+    except StopIteration as e:
+        raise StopStreaming(str(e))
+
+
+async def streaming_aiter(src, loop=None, executor=None):
```
samurang87: "aiter" sounds a bit weird... what about streaming_async_iter?

kopf/reactor/watching.py
```diff
@@ -0,0 +1,53 @@
+"""
+Watching & streaming of the watch-events.
```
samurang87: "Watching and streaming watch-events"

kopf/reactor/watching.py
```diff
@@ -0,0 +1,53 @@
...
+
+Kubernetes client's watching streams are synchronous. To make them asynchronous,
+we put them into a `concurrent.futures.ThreadPoolExecutor`,
+and yield from there asynchronously.
+
+However, async/await coroutines misbehave with `StopIteration` exceptions
+as raise by the `next` method: see `PEP-479`_.
```
samurang87: "exceptions raised by ...."

kopf/reactor/watching.py
nolar: The name is chosen after the existing convention in Python already: `__aiter__`, `__anext__`, `__aenter__`, `__aexit__` — the [magic methods](https://docs.python.org/3/reference/datamodel.html), and the ongoing duscissions all around internet about this method (e.g. [this](https://github.com/python/cpython/pull/8895) or [this](https://bugs.python.org/issue31861) or [PEP-525](https://www.python.org/dev/peps/pep-0525/#aiter-and-anext-builtins) or [google](https://www.google.com/search?q=python+anext)).

So, it seems that `aiter` is the most expected naming convention for this case.

samurang87: aah ok then

kopf/reactor/watching.py
```diff
@@ -0,0 +1,53 @@
...
+
+As a workaround, we replace `StopIteration` with our custom `StopStreaming`
+inherited from `RuntimeError` (as suggested by `PEP-479`_),
+and re-implement the generators to make them async.
+
+All of this is a workaround for the standard Kubernetes client's limitations.
+They are not needed if the client library is natively asynchronous.
```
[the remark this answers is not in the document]
nolar: Fixed.

kopf/reactor/watching.py
nolar: Fixed.

kopf/reactor/watching.py
nolar: Fixed.

Issue #13, opened 2019-04-01
Tests automation


nolar:
The project is now in the proof-of-concept stage, and lacks the tests (beside the examples, manually executed during the development).

It needs some good unit tests to freeze its current state, so that we could continue to add new features safely.

Separate the tests into the internal tests and the external promises of the interface.

Add the coverage measurements.

---

An estimation of topics to cover with the tests:

* [x] Library's public interfaces (what is exported via the main module).
* [x] #28 Object hierarchies and name/namespace/label parent-to-child propagation.
* [x] #29 Registries and `kopf.on` decorators.
* [ ] #91 (+#82) Last-seen state manipulation.
* [x] #40 Diff calculation.
* [x] #34 Handler function invocation.
* [x] #63 Event-loop running, task spawning.
* [x] #63 Watching & queueing & per-object queues/workers.
* [x] #82 Handling: cause detection and proper reaction.
* [ ] #90 (+#82) Handling: finalizers adding/removing.
* [x] #61 Handling: handler progress storage.
* [x] #35 Handling: lifecycles and handler selection.
* [ ] Sub-handler invocation and context-vars setting.
* [ ] Peering and neighbourhood awareness.
* [x] #39 CLI module loading/importing.
* [x] #39 CLI commands and options, including cluster authentication.
* [x] #71 Mocked K8s client or API calls.
* [x] #53 #54 Real-cluster execution and smoke tests.
* [x] ~#72 Coverage~ MOVED TO #99

nolar:
The remaining parts are implemented in https://github.com/nolar/kopf/pull/519.

PR #28, opened 2019-04-14
Tests for object hierarchies

nolar:
These tests "freeze" the public interface of the library regarding the hierarchy management — to prevent accidental regressions.

> Issue : #13 

Few fixes are added just to pass the tests. They slightly extend the behavior: instead of purely `list` & `tuple` classes, any iterables are accepted.

Mostly for the purpose of testing, but also as a possibility, the public interface was also extended by the namespace adjustments (move objects to the requested namespace if it is not yet set) and name harmonizing (setting either the names or the name prefixes for the objects if they are not yet set).

These two additions are directly used in the object "adoption" (i.e. assignment as a children of another parent object, and aligning all of its properties: labels, names, namespaces, etc).

### DIFF
--- a/kopf/__init__.py
+++ b/kopf/__init__.py
@@ -38,6 +38,8 @@ from kopf.reactor.queueing import (
 from kopf.structs.hierarchies import (
     adopt,
     label,
+    harmonize_naming,
+    adjust_namespace,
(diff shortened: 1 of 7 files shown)


parking52:
Hello

PR #29, opened 2019-04-14
Test registry and matching of handlers by causes

nolar:
These tests "freeze" the public interface of the library regarding the registries and matching of handlers  by causes (i.e. resources-events-fields) — to prevent accidental regressions.

> Issue : #13

### DIFF
--- a/tests/basic-structs/test_cause.py
+++ b/tests/basic-structs/test_cause.py
@@ -1,0 +1,60 @@
+import pytest
+
+from kopf.reactor.handling import Cause
+
+
+def test_no_args():
+    with pytest.raises(TypeError):
+        Cause()
+
+
+def test_all_args(mocker):
+    logger = mocker.Mock()
+    resource = mocker.Mock()
+    event = mocker.Mock()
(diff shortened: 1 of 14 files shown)


tests/basic-structs/test_cause.py
```diff
@@ -0,0 +1,60 @@
...
+
+from kopf.reactor.handling import Cause
+
+
+def test_no_args():
+    with pytest.raises(TypeError):
+        Cause()
+
+
+def test_all_args(mocker):
```
samurang87: what's the use of this test? 🤔

nolar: To test that all declared (publicly promised) args are accepted, none of them suddenly change their name, none of them are removed.

But mostly for the properties, the same: none are removed or renamed.

Both the class constructor and its properties are considered as part of the public interfaces. And the tests, by concept, should "freeze" the public promises, and they should not change when the code is changed. 

Though, maybe there are some other better approaches to do this (freeze the public interface).

Issue #32, opened 2019-04-21
Namespace-scoped operators


nolar:
## Current Behaviour

Currently, `--namespace=` option is used only to limit the watch-event to the objects belonging to that namespace, and ignoring objects in other namespaces.

However, the framework continues to use the cluster-wide watching calls, and cluster-wide peering objects. The peering objects are now cluster-scoped by definition (in CRD), and there are no namespace-scoped peering objects at all.

## Expected behaviour

As an operator developer, if I provide  `--namespace=something`, I expect that the operator limits **all** its activities to that namespace only, and does not even request for the cluster-wide objects/queries — as they can be e.g. restricted by the permissions.

If I provide `--namespace=something --peering=somepeering`, I expect that the namespace-scoped peering object kind is used, not the cluster-scoped one.

## Use-cases

**The intended use-case 1:** If an operator is a part of the application's deployment, and there are few instances of the same application deployed with different versions, but isolated by the namespaces. As a particular example: running a production, staging, and maybe experimental operators of the same application in different namespaces.

**The intended use-case 2:** Running in a cluster with strict RBAC rules, with no access to the cluster objects, restricted to one namespace only.

## Steps to Reproduce the Problem

  1. Create a RBAC service account with only the namespace permissions.
  1. Deploy any of the example operators with `--namespace=default` (or any other namespace).
  1. Observe how it fails on `api.list_cluster_custom_object() ` (in `queueing.py`/`watching.py`).
     * for `kopfexamples`
     * for `kopfpeerings`

## Acceptance Criteria

* Served objects:
  * [x] with `--namespace=`, it uses the namespace-scoped watching `api.list_namespaced_custom_object()`
  * [x] with no `--namespace=`, it uses the cluster-scoped watching `api.list_cluster_custom_object()`
* Peering objects:
  * [x] `KopfPeering` is separated to `ClusterKopfPeering` & `KopfPeering` or `NamespacedKopfPeering` (keep backward compatibility if possible, but align with the existing convention of `Role`/`ClusterRole`, etc.)
  * [x] with `--namespace=`, the namespace-scoped peering object is used.
  * [x] with no `--namespace=`, the cluster-scoped peering object is used.
  * [x] `--peering=` option only specifies the peering name, but not the listing/watching mechanics.
* Documentation:
  * [x] RBAC configuration and deployment pattern.
  * [x] CLI reference page.
  * [x] Peering page.
  * [x] A new page for cluster-vs-namespace separation ("Features" section).
[...]

Issue #31, opened 2019-04-21
Use --standalone option in  the readme of the examples


psycho-ir:
Hi, 

First of all thanks a lot for initiating this project, I think it will be a very useful project for the community. 

So about the issue, I had some difficulties to run the examples as I didn't setup the peering before jumping to the examples.  

Firstly I think it would be better to use `--standalone` parameter in the readme of the examples, so that users who haven't setup the peering yet won't get error in first run. 

Secondly I guess returning a more explicit error when the peering is not setup and referring to the relevant doc makes sense, I spent some time to debug the code and find the reason and fixed it. then I noticed it's documented here :D 
https://kopf.readthedocs.io/en/latest/install/

wdyt? I woud be more than happy to contribute to it if you think these suggestions are sensible.

nolar:
@psycho-ir Thanks for your feedback. I am glad that you like the idea of such a framework.

I think, your suggestion totally makes sense. If your suggestion is to fix the the docs & `./examples/*/README.md` files only, feel free to do it — I welcome your contribution.

If you want to change the framework behaviour, please take a look at  #32 & #33 first — I have just written down some of my thoughts on this topic.

The only thing that stops me from quickly implementing this, is the absence of tests. The framework has just recently left the proof-of-concept stage, and I cover it with the tests component-by-component, module-by-module (not fast enough, sorry).

psycho-ir:
I just read them, yep they make totally sense to me. 
I think by implementing #33 we don't even to change the doc and adding the `--standalone` to the docs. 

I can work on it this week, or you would rather it's implemented after #32? to have support of the namespaced peerings in the first run.

nolar:
@psycho-ir I would say they are independent (unless there are some tricky details which I do not foresee). 

You can just keep the always-cluster-scoped concept for now, and implement the logic for the present/absent peering object.

nolar:
@psycho-ir Please, take a look at #36 & #37 — regarding the #32 (namespace isolation of the served objects and of the peering objects). Seems working as expected — I will double-check tomorrow, with fresh mind.

You can send your implementation for #33 as is — I will rebase&resolve these two PRs later (they are less important than a working tutorial).

nolar:
Indirectly fixed by #33 & #38 (auto-detection of the peering mode) — the tutorial does not break if the cluster is not configured with the peering object.

Issue #17, opened 2019-04-02
Add the RBAC examples for deployments


nolar:
Kopf is just a framework, but the Kopf-based operators must be deployed to the cluster. For that, they would need the RBAC (role-based access control) templates and examples.

Add and document some common templates with the RBAC objects (roles, rolebindings, etc).

nolar:
The RBAC generation idea is extracted to #49. This issue was originally about the docs only, and it was done some time ago: https://kopf.readthedocs.io/en/stable/deployment/#rbac

Issue #15, opened 2019-04-02
Consider pykube-ng?


nolar:
Originally by @hjacobs :

> I see that you are currently using the "official" Kubernetes client (Swagger Codegen).
> I forked the old pykube to https://github.com/hjacobs/pykube as I'm rather unhappy about the complexity and size of the official Kubernetes client.
> See also https://github.com/hjacobs/pykube/issues/12
> Not sure if this would even work or whether you use something specific of the Kubernetes Python client.

nolar:
After #71, almost all Kubernetes-related code is consolidated in one package: `kopf.k8s`, where it is easy to be replaced by any other implementation. The only thing outside of this package is authentication (`kopf.config.login()`).

However, the whole codebase of Kopf assumes that the objects manipulated as dicts. Not even the Kubernetes client's "models". A lot of `.get('metadata', {}).get('something')` and `.setdefault('status', {}).setdefault('kopf', {})` and similar lines are all around. It will be difficult to change that and to support both dicts & client's classes. It is better not to do so. 

In addition, Kopf promises to hide implementation details from the user ⟹ which means that the user should not know which Kubernetes client is used under the hood ⟹ which means that the internal models/classes must not be exposed ⟹ which means they have to be converted to dicts on arrival.

---

Another tricky part will be watch-streaming. The official Kubernetes client does that in the `while True` cycle, and reconnects all the time. Pykube-ng exits after the first disconnection, which happens roughly ~5s, or as specified by a timeout query arg. The connection cannot stay forever, so either the client, or Kopf should handle the reconnections. Some partial workarounds can be found in #96 (pre-listing and `resourceVersion` usage).

nolar:
So far, a list of issues detected while trying to switch to pykube:

* `pykube.HTTPClient` object is needed on every call, there is no implicit config (as in the official client). Has to be stored globally and reused all over the code.
* `pykube.HTTPClient` has a default timeout of 10s for any connection, including the watching. Can be overridden explicitly with `timeout=None`, but requires a separate `pykube.HTTPClient` instance for that.
* `pykube.HTTPClient` raises exceptions from `requests` on timeouts, not its own.
* Watch-call terminates after the connection is lost for any reason, no internal reconnection or `while True`. Has to be caught and repeated. In the official K8s client, it is done internally: the watch is eternal.
* `object_factory()` prints a list of discovered resources to stdout, this is a visual garbage.
[...]

PR #36, opened 2019-04-22
Peering scopes

nolar:
> Issue : #32, related #31, #17

For the purpose of strict namespace isolation of the operators, the peering objects must be cluster- or namespace-scoped (previously: always cluster-scoped). 

For this, split the old cluster-wide `KopfPeering` resource into the new `ClusterKopfPeering` & namespaced `KopfPeering`, and use one of them depending on the `--namespace` option.

The testing of peering is performed manually — it works. More tests will be added when the whole peering subsystem will be covered with tests (as part of #13).

Docs preview:

* https://kopf.readthedocs.io/en/peering-scopes/peering/
* https://kopf.readthedocs.io/en/peering-scopes/deployment/#rbac

See also: #37 for the namespace isolation of the listing/watching API calls.

TODO:

* [x] Manually test how do the operators react to changes in the `kind: KopfPeering` when re-created from the cluster to the namespaced scope — in different running modes (with/without `--peering`, with/without `--namespace`). Ensure that the operators actually behave as before and as expected.
* [x] Fallback to the old CRD `KopfPeering` if that CRD is cluster-scoped (but not if namespaced).
* [x] Document the upgrade scenario to code first, CRDs second, never vice versa.

### DIFF
--- a/docs/deployment.rst
+++ b/docs/deployment.rst
@@ -78,7 +78,7 @@ and to manipulate the objects, both domain-specific and the built-in ones.
 For the example operator, those are:
 
 * ``kind: ClusterKopfPeering`` for the cross-operator awareness (cluster-wide).
-* ``kind: NamespacedKopfPeering`` for the cross-operator awareness (namespace-wide).
+* ``kind: KopfPeering`` for the cross-operator awareness (namespace-wide).
 * ``kind: KopfExample`` for the example operator objects.
 * ``kind: Pod/CluJob/PersistentVolumeClaim`` as the children objects.
 * And others as needed.
--- a/docs/deployment.rst
+++ b/docs/deployment.rst
@@ -77,9 +77,10 @@ The pod where the operator runs must have the permissions to access
 and to manipulate the objects, both domain-specific and the built-in ones.
 For the example operator, those are:
 
-* ``kind: KopfPeering`` for the cross-operator awareness.
+* ``kind: ClusterKopfPeering`` for the cross-operator awareness (cluster-wide).
+* ``kind: NamespacedKopfPeering`` for the cross-operator awareness (namespace-wide).
 * ``kind: KopfExample`` for the example operator objects.
-* ``kind: Pod/Job/PersistentVolumeClaim`` as the children objects.
+* ``kind: Pod/CluJob/PersistentVolumeClaim`` as the children objects.
 * And others as needed.
 
 For that, the RBAC_ (Role-Based Access Control) could be used
@@ -87,7 +88,9 @@ and attached to the operator's pod via a service account.
 
 .. _RBAC: https://kubernetes.io/docs/reference/access-authn-authz/rbac/
 
-Here is an example of what a RBAC config should look like:
+Here is an example of what a RBAC config should look like
+(remove the parts which are not needed: e.g. the cluster roles/bindings
+for the strictly namespace-bound operator):
 
 .. literalinclude:: deployment-rbac.yaml
     :caption: rbac.yaml
@@ -104,7 +107,7 @@ And the created service account is attached to the pods as follows:
     :name: deployment-service-account-yaml
 
 
-The service accounts are always namespace-scoped.
+Please note that the service accounts are always namespace-scoped.
 There are no cluster-wide service accounts.
 They must be created in the same namespace as the operator is going to run in
-(even if it is going to server the whole cluster).
+(even if it is going to serve the whole cluster).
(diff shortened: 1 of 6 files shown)


kopf/reactor/peering.py
```diff
@@ -100,34 +102,46 @@ def keepalive(self):
...
+    body = {'status': {peer.id: None if peer.is_dead else peer.as_dict() for peer in peers}}
     api = kubernetes.client.CustomObjectsApi()
-    api.patch_cluster_custom_object(
-        group=PEERING_CRD_RESOURCE.group,
-        version=PEERING_CRD_RESOURCE.version,
-        plural=PEERING_CRD_RESOURCE.plural,
-        name=peering,
-        body={'status': {peer.id: None if peer.is_dead else peer.as_dict() for peer in peers}},
-    )
+    if namespace is None:
```
psycho-ir: what about wrapping this logic in a class? so we can use it later to fetch/patch peering custom objects.
 e.g.
```
PeeringObjects.patch(body, peering, namespace: Optional[String]) 
PeeringObjects.exists(peering, namespace: Optional[String]) -> bool
```

nolar: [`pykube-ng`](https://github.com/hjacobs/pykube) does that, more or less. I prefer to not re-invent it or any other Kubernetes client. Instead, it is better to built on top of pykube-ng (see #15). I have a draft branch, but there are few issues both in pykube-ng, and in pykube-to-kopf connectivity — I am going to get back to them later, after I cover the functionality with the tests.

nolar:
So far, as tested with Minikube, the behaviour is this:

* **Baseline:** Old code, old peerings (KopfPeering is cluster-scoped, ClusterKopfPeering does not exist):
  * No CLI options: Auto-detection mode, cluster-wide peering was used.
  * `--namespace=default`: Uses the cluster-wide peering.
  * `--peering=default`: Uses the cluster-wide peering.
  * `--namespace=default --peering=default`: Uses the cluster-wide peering.
* **Mainline:** New code, old peerings (KopfPeering is cluster-scoped, ClusterKopfPeering does not exist):
  * No CLI options: Auto-switches to standalone mode with a warning (mismatches❗️).
  * `--namespace=default`: Uses the cluster-wide peering (as expected ✅ ).
  * `--peering=default`: Fails with "The peering was not found (mismatches❗️).
  * `--namespace=default --peering=default`: Uses the cluster-wide peering (as expected ✅ ).
* **Edgecase:** Old code, new peerings (KopfPeering is namespaced, ClusterKopfPeering is cluster-scoped).
  * Fails with "Namespace parameter required" (mismatches❗️).

---

Both cases of (new code + old peerings) fail because they try to use the `ClusterKopfPeering` CRD, which does not exist there — due to `if namespace is None` condition. They work fine if the CRD is the old one, `KopfPeering`.

---

All cases of (old code + new peerings) fail since they use the now-namespaced CRD (KopfPeering) with a cluster-scoped method (`get_cluster_custom_object()`).

This is not solvable, since it can only be done with a new version which supports the fallback to `ClusterKopfPeering` — and if one needs to upgrade, why not to upgrade to the new code with both old & new peerings supported; i.e. the transitional releases make no sense.

---

TODO:

* [ ] Fallback to the old CRD `KopfPeering` if that CRD is cluster-scoped (but not if namespaced).
* [ ] Document the upgrade scenario to code first, CRDs second, never vice versa.

nolar:
Rebased on master, and added the fallback scenario for the legacy mode (cluster-scoped KopfPeering). As tested manually, the new code works as expected both in the legacy mode (of the cluster) and in the new mode (as documented).

@samurang87 @aweller It can be reviewed and merged to master now.

docs/deployment.rst
```diff
@@ -77,17 +77,20 @@ The pod where the operator runs must have the permissions to access
 and to manipulate the objects, both domain-specific and the built-in ones.
 For the example operator, those are:
 
-* ``kind: KopfPeering`` for the cross-operator awareness.
+* ``kind: ClusterKopfPeering`` for the cross-operator awareness (cluster-wide).
+* ``kind: KopfPeering`` for the cross-operator awareness (namespace-wide).
 * ``kind: KopfExample`` for the example operator objects.
-* ``kind: Pod/Job/PersistentVolumeClaim`` as the children objects.
+* ``kind: Pod/CluJob/PersistentVolumeClaim`` as the children objects.
```
samurang87: Clujob?

docs/peering.rst
```diff
@@ -3,28 +3,150 @@ Peering
...
+And so on.
+
+For this, assign a monotonically growing or random priority to each
+operator in the deployment or replicaset:
+
+.. code-block:: bash
+
+    kopf run --priority=$RANDOM ...
+
+``$RANDOM`` is a feature of bash (if you use another shell, find your own way).
```
samurang87: "find your own way" is kinda rude :P

docs/deployment.rst
nolar: Fixed.

docs/peering.rst
nolar: Fixed.

PR #37, opened 2019-04-22
Restrict the API calls in namespaced mode

nolar:
> Issue : #32, related #31 

Reimplement the namespace filtering of `--namespace` option to use the appropriate API calls instead of object field checking. 

First, this will remove the unnecessary receiving and silent ignoring of the irrelevant objects from other namespaces. 

Second, this can simplify the RBAC configs for the per-namespace operators (no cluster calls are made, thus no cluster roles are needed — see docs preview: https://kopf.readthedocs.io/en/peering-scopes/deployment/#rbac).

The namespace separation of regular objects is tested both manually and with limited set of unit-tests. More tests will be added when the whole watching-queueing-handling subsystem will be covered with tests (as part of #13).

See also #36 for the namespace isolation of the peering objects.


kopf/reactor/watching.py
```diff
@@ -51,3 +62,68 @@ def streaming_next(src):
             yield await loop.run_in_executor(executor, streaming_next, src)
         except StopStreaming:
             return
+
+
+async def streaming_watch(
+        resource: Resource,
+        namespace: Union[None, str],
```
samurang87: interesting

Issue #33, opened 2019-04-21
Auto-guessing the peering mode


nolar:
## Current Behaviour

Currently, the peering object is needed by default, unless `--standalone` option is used, which disables the peering completely.

This causes the confusion for the first intro and following the tutorial — in case the cluster is not configured yet (no peering objects created). See: #31.

If standalone mode is made the default, there is a negative side-effect: if somebody runs 2+ operators —e.g. one in-cluster, another in the dev-mode on the external workstations— these operators will collide and compete for the objects without knowing this. The peering was invented exactly for the purpose of not hitting this issue in the dev-mode, and gracefully "suppressing" other operators.

## Expected Behaviour

The peering should be considered as a side-feature for extra safety, it should not be a showstopper for the quick-start guides or tutorials.

It would be better to have 3 modes:

* with ~`--peering` or~ `--peering=something`, the peering is enforced, the operator fails to start if peering is not accessible (as it is now).
* with `--standalone`, the peering is ignored (as it is now).
* with no options (the new default), the auto-detection mode is used: if the "metadata.name: default" peering object is found, use it (either cluster-scoped or namespace-scoped, depending on `--namespace=`); if not found, log a big-letter warning of possible conflicts and collisions, and continue as if in the standalone mode.

Relevant: #32.

## Todos:

* [x] Documentation:
  * [x] CLI options.
  * [x] Peering page.
* [ ] Tests.

psycho-ir:
@nolar I think it's better to not support `--peering` as flag. It might bring ambiguity in cases like this:
`run ../examples/01-minimal/example.py --peering --verbose`
It can be interpreted as a peering with name `--verbose`. 
IMO it's safe to switch to auto-mode if `--peering` does not exist and just show the warning to the user. 
wdyt?

nolar:
> I think it's better to not support `--peering` as flag.

@psycho-ir  Agree. Also, it is unclear what is meant if it is just a flag: which peering object to use.

I've fixed the issue text.

Issue #30, opened 2019-04-18
Silent handlers (spies)


nolar:
Currently, handlers are fully orchestrated by the framework, and their progress is stored in the object's `status.kopf` field. This limits the handlers only to the objects designed specially for this operator.

Sometimes, it is needed to watch after objects of different kind, or maybe even the built-in objects, such as pods, jobs, so on — e.g. created as the children of the main object. Putting the Kopf's progress field on these objects causes few problems:

* Unnecessary changes & watch-events on those objects, as noticed by other controllers and operators.
* Multiple Kopf-based operators will conflict with each other, as they use the same status subfield (see #23).

Kopf should provide a way to watch for the built-in objects _**silently**_:

```
import kopf

@kopf.on.event('', 'v1', 'pods')
def pod_changed(body, **kwargs):
    pass
```

This induced few limitations:

* Progress is reported only in the logs, maybe in the k8s-events, but not in the status fields.
* If the handler fails, there will be no retries as with the normal handlers. It can even miss the actual change and needed reaction in that case until the next event happens.
* No cause detection will be provided (i.e. field diffs detection), only the raw events as they are sent by k8s.

This functionality is already present in the Kopf's reactor (the stream between the queueing & handling modules), so it makes sense to expose it as a feature.

---

Also, once done, add the missing docs for the [sample problem](https://kopf.readthedocs.io/en/stable/walkthrough/problem/): it should track when and if the PVCs are bound and finally unbound — so that the tutorial is indeed full and complete.

---

* [ ] Silent handlers implemented.
* [ ] Tests.
* [ ] Docs on the feature in the "Handlers" section.
* [ ] Tutorial extended for the PVC monitoring for being bound, and activating the deletion afterwards.

PR #38, opened 2019-04-22
[GH-33] implement auto detection mode for peering.

psycho-ir:
> Issue : #33 (only if appropriate)

@nolar I will write the tests and document it once we agreed on the implementation.

### DIFF
--- a/docs/peering.rst
+++ b/docs/peering.rst
@@ -11,9 +11,10 @@ The operator can be instructed to use alternative peering objects::
 
 The operators from different peering objects do not see each other.
 
-The default peering name (i.e. if no peering or standalone options are provided)
-is ``default``.
+Default behavior
+----------------
 
+If there is a peering object with name `default` then it's been used by default as the peering object. Otherwise kopf will run the operator in mode `Standalone`.
 
 Standalone mode
 ---------------
--- a/kopf/cli.py
+++ b/kopf/cli.py
@@ -87,6 +87,3 @@ def resume(id, namespace, peering):
         namespace=namespace,
     )
     ourselves.disappear()
-
-
-main()
--- a/kopf/cli.py
+++ b/kopf/cli.py
@@ -32,7 +32,7 @@ def main():
 @click.option('-n', '--namespace', default=None)
 @click.option('--standalone', is_flag=True, default=False)
 @click.option('--dev', 'priority', flag_value=666)
-@click.option('-P', '--peering', type=str, default=PEERING_DEFAULT_NAME)
+@click.option('-P', '--peering', type=str, default=None)
 @click.option('-p', '--priority', type=int, default=0)
 @click.option('-m', '--module', 'modules', multiple=True)
 @click.argument('paths', nargs=-1)
@@ -87,3 +87,6 @@ def resume(id, namespace, peering):
         namespace=namespace,
     )
     ourselves.disappear()
+
+
+main()
--- a/kopf/reactor/peering.py
+++ b/kopf/reactor/peering.py
@@ -79,6 +79,26 @@ class Peer:
     def __repr__(self):
         return f"{self.__class__.__name__}({self.id}, namespace={self.namespace}, priority={self.priority}, lastseen={self.lastseen}, lifetime={self.lifetime})"
 
+    @classmethod
+    def detect(cls,
+               standalone: bool,
+               peering: Optional[str],
+               **kwargs) -> Optional:
+        if standalone:
+            return None
+
+        if peering:
+            if Peer._is_peering_exist(peering):
+                return cls(peering=peering, **kwargs)
+            else:
+                raise Exception(f"The peering {peering} was not found")
+
+        if Peer._is_default_peering_setup():
+            return cls(peering=peering, **kwargs)
+
+        logger.warning(f"The default peering object not found. Falling back to the Standalone mode...")
+        return None
+
     def as_dict(self):
         # Only the non-calculated and non-identifying fields.
         return {
@@ -109,6 +129,24 @@ class Peer:
         """
         self.touch(lifetime=0)
         apply_peers([self], peering=self.peering)
+
+    @staticmethod
+    def _is_default_peering_setup():
+        return Peer._is_peering_exist(PEERING_DEFAULT_NAME)
+
+    @staticmethod
+    def _is_peering_exist(peering: str):
+        api = kubernetes.client.CustomObjectsApi()
+        try:
+            api.get_cluster_custom_object(group=PEERING_CRD_RESOURCE.group,
+                                          version=PEERING_CRD_RESOURCE.version,
+                                          plural=PEERING_CRD_RESOURCE.plural,
+                                          name=peering)
+            return True
+        except ApiException as e:
+            if e.status == 404:
+                return False
+            raise
 
 
 def apply_peers(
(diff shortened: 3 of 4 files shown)


kopf/reactor/peering.py
```diff
@@ -190,7 +191,7 @@ def apply_peers(
 
             # How often do we update. Keep limited to avoid k8s api flooding.
             # Should be slightly less than the lifetime, enough for a patch request to finish.
-            await asyncio.sleep(max(1, ourselves.lifetime.total_seconds()-10))
+            await asyncio.sleep(max(1, int(ourselves.lifetime.total_seconds() - 10)))
```
nolar: What was the problem with this? In theory, it can be float, and I believe there is nothing wrong with it (except that in practice, updating the peering object too often is not a good idea).

kopf/reactor/queueing.py
```diff
@@ -183,13 +183,8 @@ def create_tasks(
...
 
     # Monitor the peers, unless explicitly disabled.
-    if not standalone:
-        ourselves = Peer(
-            id=detect_own_id(),
-            priority=priority,
-            peering=peering,
-            namespace=namespace,
-        )
+    ourselves: Optional[Peer] = PeerFactory.create_peer(standalone, peering, id=detect_own_id(), priority=priority, namespace=namespace)
```
nolar: I didn't know it is syntactically allowed to do so in the code flow. It is nice to learn something new :-)

kopf/reactor/peering.py
```diff
@@ -227,3 +228,43 @@ def detect_own_id() -> str:
     now = datetime.datetime.utcnow().isoformat()
     rnd = ''.join(random.choices('abcdefhijklmnopqrstuvwxyz0123456789', k=6))
     return f'{user}@{host}/{now}/{rnd}'
+
+
+class PeerFactory:
```
nolar: What I would suggest, is to consider using the `@classmethod` in `Peer` instead of creating the `PeerFactory` entity with `@staticmethods` (since any class is already a factory): e.g. `ourselves = Peer.detect(...)` — which can also return `None`. 

But since there is no hierarchy of peering classes, this is not critical.

kopf/cli.py
```diff
@@ -87,3 +87,6 @@ def resume(id, namespace, peering):
         namespace=namespace,
     )
     ourselves.disappear()
+
+
+main()
```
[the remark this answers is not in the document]
psycho-ir: Ah right, it was not supposed to be committed. will remove it.

kopf/reactor/peering.py
psycho-ir: just to get rid of the type checking warning of the pycharm. as  `max` function  accepts (int,int) or iterable. otherwise there was no problem with code itself.

kopf/reactor/queueing.py
psycho-ir: it's introduced in python3.7, quite new :D

kopf/reactor/peering.py
psycho-ir: Good point, done.

kopf/reactor/peering.py
```diff
@@ -79,6 +79,26 @@ def __init__(self,
...
+            return None
+
+        if peering:
+            if Peer._is_peering_exist(peering):
+                return cls(peering=peering, **kwargs)
+            else:
+                raise Exception(f"The peering {peering} was not found")
+
+        if Peer._is_default_peering_setup():
+            return cls(peering=peering, **kwargs)
```
nolar: Here, it causes a crash on start in auto-detection mode (when `peering is None`). Perhaps, this was intended:

```python
return cls(peering=PEERING_DEFAULT_NAME, **kwargs)
```

Otherwise, it passes `None`, and then it remembers `ourselves.peering=None`, and cannot perform `PATCH` on `None`, and fails on start in the auto-detection mode.

nolar:
@psycho-ir PS: Also, feel free to remove the clutter from the PR body if you want. (I think, the PR template must be reduced to only an issue reference (it is mandatory), with no implicit structure. — Will do later.)

kopf/reactor/peering.py
nolar: I see. Interesting that it wants the integers only. Anyway, the rounding does not affect the behaviour significantly.

kopf/reactor/queueing.py
nolar: Oh! Good that I've started this project with `>=3.7` requirement, not `>=3.6`! :-) (Also, some 3.7-only ISO datetime functions are used.) — Absence of legacy is always good.

psycho-ir:
> I've tried it locally. There is a little bug with the `...cls(peering=None,...` — crashes on start (see comments). Once fixed, it works nice in all modes, exactly as intended.
> 
> Would you like to extend the docs (`docs/peering.rst` → https://kopf.readthedocs.io/en/latest/peering/) in this PR? Or I can do this in the following PRs, together with other peering doc changes.

Sure, I will extend the docs tomorrow and try to add some tests for it.

kopf/reactor/peering.py
psycho-ir: Ah true, I didn't test this path manually 👍 
merged.

docs/peering.rst
```diff
@@ -11,9 +11,10 @@ The operator can be instructed to use alternative peering objects::
 
 The operators from different peering objects do not see each other.
```
psycho-ir: @nolar I updated the peering doc.

kopf/reactor/peering.py
```diff
@@ -78,6 +79,26 @@ def __init__(self,
...
+        if peering:
+            if Peer._is_peering_exist(peering):
+                return cls(peering=peering, **kwargs)
+            else:
+                raise Exception(f"The peering {peering} was not found")
+
+        if Peer._is_default_peering_setup():
+            return cls(peering=PEERING_DEFAULT_NAME, **kwargs)
+
+        logger.warning(f"The default peering object not found. Falling back to the Standalone mode...")
```
samurang87: * Default peering object not found

psycho-ir:
```suggestion
        logger.warning(f"Default peering object not found, falling back to the Standalone mode.")
```

nolar:
@psycho-ir So, it is merged and released as `kopf==0.9`. Congratulation! And big thanks for your contribution!

psycho-ir:
@nolar yaaaaaay. what is the next milestone? Would be happy to contribute more in this project.

nolar:
@psycho-ir 

Currently, the milestone 1 is this:

* Tests, tests, tests — to bring the repo to a healthy state, so that I am not afraid to introduce new changes without breaking things. 90% of them are done, just in the PRs or in my local branches waiting for some PRs to be merged.

* Silent spies on the events (see #30 ) — to react to the events in pods, persistent volume claims, etc, without storing the handler status. Already implemented in my local branch, waiting for the tests.

* Finish the tutorial in the docs, so that the `kind: EphemeralVolumeClaim` becomes a real example operator in its own repo, uploaded to the DockerHub, etc. Partially drafted in the docs (in pieces), though not actually tested in real cluster. Waiting for the missing feature of the silent spies.

---

Since then, the framework is sufficiently feature-rich for the first stage (it is now, actually, just the docs do not feel complete), and can be advertised in public: meetups, blog posts, so on. 

Based on the real-world feedback, the next milestones can be defined.

The real-world usage is the most important goal now. I.e., getting the operators implemented with this framework (and preferably shared).

Meanwhile, I write down all the ideas that come to my mind as the issues. If you have some suggestions, feel free to create the issues too. Examples: #44, #45, #46.

PR #40, opened 2019-04-22
Test diffs calculation and manipulation

nolar:
> Issue : #13 

Diffs are used in the `kopf.on.update` handlers, and then in the per-field handlers. They are very minimalistic and very simple structures, easy to test.

### DIFF
--- a/tests/diffs/test_diff.py
+++ b/tests/diffs/test_diff.py
@@ -1,0 +1,78 @@
+from kopf.structs.diffs import diff
+
+
+def test_scalars_equal():
+    a = 100
+    b = 100
+    d = diff(a, b)
+    assert d == ()
+
+
+def test_scalars_unequal():
+    a = 100
+    b = 200
+    d = diff(a, b)
+    assert d == (('change', (), 100, 200),)
+
+
+def test_strings_equal():
+    a = 'hello'
+    b = 'hello'
+    d = diff(a, b)
+    assert d == ()
+
+
+def test_strings_unequal():
+    a = 'hello'
+    b = 'world'
+    d = diff(a, b)
+    assert d == (('change', (), 'hello', 'world'),)
+
+
+def test_lists_equal():
+    a = [100, 200, 300]
+    b = [100, 200, 300]
+    d = diff(a, b)
+    assert d == ()
+
+
+def test_lists_unequal():
+    a = [100, 200, 300]
+    b = [100, 666, 300]
+    d = diff(a, b)
+    assert d == (('change', (), [100, 200, 300], [100, 666, 300]),)
+
+
+def test_dicts_equal():
+    a = {'hello': 'world', 'key': 'val'}
+    b = {'key': 'val', 'hello': 'world'}
+    d = diff(a, b)
+    assert d == ()
+
+
+def test_dicts_with_keys_added():
+    a = {'hello': 'world'}
+    b = {'hello': 'world', 'key': 'val'}
+    d = diff(a, b)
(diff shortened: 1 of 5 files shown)


tests/diffs/test_diff.py
```diff
@@ -0,0 +1,78 @@
...
+    b = {'hello': 'world', 'key': 'new'}
+    d = diff(a, b)
+    assert d == (('change', ('key',), 'old', 'new'),)
+
+
+def test_dicts_with_subkeys_changed():
+    a = {'main': {'hello': 'world', 'key': 'old'}}
+    b = {'main': {'hello': 'world', 'key': 'new'}}
+    d = diff(a, b)
+    assert d == (('change', ('main', 'key'), 'old', 'new'),)
```
parking52: I would like to have a test with a real usecase with Dicts that the reactor is using; which would get closer to 'test as documentation'

```suggestion
def test_dicts_sample_kopf_usecase_update():
    body_before_update = {'metadata' : {'annotations' : 'abcdef'}}
    body_after_update = {'metadata' : {'annotations' : 'abcdef'}}
```

nolar: @parking52 Good idea. 

Though, the `resourceVersion` will never be in the diffs, as this and other system fields are filtered out of the last-seen state calculation (see [`kopf.structs.lastseen.get_state()`](https://github.com/zalando-incubator/kopf/blob/fa4d77244dd1dd0e95f89b0f553173649d11db04/kopf/structs/lastseen.py#L30-L45)) — for the reason that these fields change even when we do not care about these changes. 

Only all of the the `spec` fields & some (non-filtered) `metainfo` fields, such as `metainfo.labels`, `metainfo.annotations` (except few specific annotations) will be taken into account, diffs calculated, and the update-handlers called.

Maybe, you can implement it for all of the mentioned above? I.e.:

* metainfo.labels.some-label
* metainfo.annotations.some-annotation
* spec.field (a string)
* spec.items (a list of strings)
* spec.modes (a dict of strings=>strings).

These are the main real-world examples happening in the operators and expected in `@kopf.on.field()`. 

See also: [`examples/obj.yaml`](https://github.com/zalando-incubator/kopf/blob/fa4d77244dd1dd0e95f89b0f553173649d11db04/examples/obj.yaml), which I use(d) for the local runs and manual testing.

tests/diffs/test_resolving.py
```diff
@@ -0,0 +1,28 @@
...
+def test_unexisting_key():
+    d = {'abc': {'def': {'hij': 'val'}}}
+    with pytest.raises(KeyError):
+        resolve(d, ['abc', 'def', 'xyz'])
+
+
+def test_nonmapping_key():
+    d = {'key': 'val'}
+    with pytest.raises(TypeError):
+        resolve(d, ['key', 'sub'])
```
parking52:
```suggestion
        resolve(d, ['key', 'val'])
```
I understand from non mapping that the key must be correct though

nolar: The unexistent key is tested in `test_unexisting_key` above. There, the data structure before failure is a mapping `d['abc']['def']` equal to `{'hij': 'val'}`, and we tried `{'hij': 'val'}['xyz']` — it raises `KeyError`.

Here, we use the actually existing key — `d['key']` — which has a value that is not a mapping: `"val"`. It is not possible to do `"val"['key']` — it raises `TypeError`. 

I can probably align this test with the previous one so that they look the same, and then I can select `['abc']['def']['hij']['key']` — with the same effect.

Issue #42, opened 2019-04-26
Ignore the events from the past


nolar:
## Actual Behaviour

Kopf-based operator reacts to the object creation events in some cases. In more details, it is described in https://github.com/kubernetes-client/python/issues/819.

**Briefly:** it is caused by how a kubernetes client library is implemented: it remembers the last seen resource version among all objects as they are listed on the initial call. Kubernetes lists them in arbitrary order, so the old ones can be the latest in the list. Then, the client library uses that old resource version to re-establish the watch connection, which replays all the old events since that moment in time when this resource version was the latest. This also includes the creation, modification, and even the deletion events for the objects that do not exist anymore.

In practice, it means that the operator will call the handlers, which can potentially create the children objects and do some other side effects. In our case, it happened every day when some cluster events were executed.; but it could happen any time the existing watch connection is re-established.

## Expected Behaviour

The operator framework should follow the "eventual consistency" principle, which means that only the last state (the latest resource version, the latest event) should be handled.

Since the events are streaming, the "batch of events" can be defined as a time-window of e.g. 0.1s — fast enough to not delay the reaction in normal cases, but slow enough to process all events happening in a row.

## Steps to Reproduce the Problem

Create some amount (10-20) of objects.

Example for my custom resource kind:

```
In [59]: kubernetes.config.load_kube_config()  # developer's config files
In [60]: api = kubernetes.client.CustomObjectsApi()
In [61]: api_fn = api.list_cluster_custom_object
In [62]: w = kubernetes.watch.Watch()
In [63]: stream = w.stream(api_fn, 'example.com', 'v1', 'mycrds')
In [64]: for ev in stream: print((ev['type'], ev['object'].get('metadata', {}).get('name'), ev['object'].get('metadata', {}).get('resourceVersion'), ev['object'] if ev['type'] == 'ERROR' else None))

('ADDED', 'mycrd-20190328073027', '213646032', None)
('ADDED', 'mycrd-20190404073027', '222002640', None)
('ADDED', 'mycrd-20190408065731', '222002770', None)
('ADDED', 'mycrd-20190409073007', '222002799', None)
('ADDED', 'mycrd-20190410073012', '222070110', None)
('ADDED', 'mycrd-20190412073005', '223458915', None)
('ADDED', 'mycrd-20190416073028', '226128256', None)
('ADDED', 'mycrd-20190314165455', '233262799', None)
('ADDED', 'mycrd-20190315073002', '205552290', None)
[...]

PR #43, opened 2019-04-26
Handle only the latest event with the most up to date state

nolar:
> Issue : #42 

With this change, if the queue contains a batch of events, react only to the latest one — to follow the "eventual consistency" principle of Kubernetes.

The batch is time-framed to 0.1s, so that it is fast enough for all normal cases when only one event arrives, but slow enough for the initial object listing when multiple events arrive. 

Still, some minimal time-frame is needed, as the streaming and parsing of the events inside of the kubernetes library is not immediate.

### DIFF
--- a/kopf/reactor/queueing.py
+++ b/kopf/reactor/queueing.py
@@ -140,10 +140,17 @@ async def worker(
 
             # Try ASAP, but give it few seconds for the new events to arrive, maybe.
             # If the queue is empty for some time, then indeed finish the object's worker.
+            # If the queue is filled, use the latest event only (within the short timeframe).
             try:
                 event = await asyncio.wait_for(queue.get(), timeout=5.0)
             except asyncio.TimeoutError:
                 break
+            else:
+                try:
+                    while True:
+                        event = await asyncio.wait_for(queue.get(), timeout=0.1)
+                except asyncio.TimeoutError:
+                    pass
 
             # Try the handler. In case of errors, show the error, but continue the queue processing.
             try:


samurang87:
So basically if the fetching manages to get any events, you restart it to make sure there is only one event fetched? 🤔

nolar:
@samurang87 Yes. And getting all the events till nothing is left. All events except the last one are therefore ignored — as intended.

parking52:
Why will the event we are interested in be fetched rapidly ? And why will the number of event decrease ?

samurang87:
I'm confus 🤔

nolar:
@parking52 @samurang87  I didn't get your last comments. What number of events decreased? What exactly is fetched rapidly?

nolar:
@parking52 @samurang87 See the examples in #42 — there are few objects with few events (e.g. "mycrd-expr1"), of which only the latest is of our interest.

nolar:
@parking52 @samurang87 This is efficiently an equivalent of this logic:

```python
while queue.qsize() > 0:
    event = await queue.get()
```

Except that:

* `threading.Queue().qsize()` on Mac OS (it is not supported), so I used to avoid `qsize()` and just to not think about it. It is promised to work normally in `asyncio.Queue` though.
* If the queue is empty (qsize == 0), we should wait for 5.0 seconds until something appears. So, `while queue.qsize() > 0:` is not a suitable criterion here, unless `first_time` is introduced. But it makes it same as not nice as double tries. In that case, it would look like this:

```python
event = None
first_time = True
try:
    while first_time or queue.qsize() > 0:
        first_time = False
        event = await asyncio.wait_for(queue.get(), timeout=5.0)
except asyncio.TimeoutError:
    if event is None:
        break

# when (event is not None) or (no timeout on the last event), continue.
```

Issue #49, opened 2019-04-29
Automated RBAC generation and verification


nolar:
## Background

With `kopf>=0.9`, the operators fail to start in the clusters with RBAC configured according to the docs. Introduced by #38, where `GET` is used on a specific peering object (not just on a list).

The deployment docs were not updated to reflect that. And, even if updated, that would lead to these incidents anyway, as the RBAC yaml file is not auto-validated and not auto-updated in our case, so we would not notice the change.

## Suggestion: RBAC verification

Kopf should allow to verify if the RBAC yaml file matches the framework's and operator's expectations, and explain what is missing:

```bash
kopf rbac verify script1.py script2.py -f rbac.yaml
```

This verification step could be optionally used either in CI/CD testing stage, or in the docker build stage, and to fail the build if the source-code RBAC yaml file lacks some necessary permissions.

If no `-f` option is specified (OR: if `--cluster` is explicitly specified — TDB), then verify against the real currently authenticated cluster:

```bash
kopf rbac verify script1.py script2.py --cluster
```

The output should explain what is missing:

```
# Kopf's internals:
KopfPeering get permission: ❌absent
KopfPeering list permission: ✅present
KopfPeering watch permission: ✅present
KopfPeering patch permission: ✅present

# Used at script1.py::create_fn():
KopfExample list permission: ✅present
KopfExample watch permission: ✅present
KopfExample patch permission: ✅present

Some permissions are missing. The operator will fail to work.
Read more at https://kopf.readthedocs.io/en/stable/deployment/
Or use `kopf rbac generate --help`
```

Exit status should be 0 (all is okay) or 1 (something is missing), so that it could be used in CI/CD.

## Suggestion: RBAC generation

Since Kopf would already contain the RBAC parsing & analysis logic, it is also fine to generate the RBAC yaml files from the codebase of the operator — based on which resources/events/causes are registered for handling (same CLI semantics as in `kopf run`: `-m module` or `file.py`).

```
kopf rbac generate script1.py script2.py > rbac.yaml
kubectl apply -f rbac.yaml
```

## Extra: children objects introspection

As a challenge, some introspection might be needed into the internals of the handlers on which children objects they manipulate from the handlers (e.g. pod creation) — this must also be part of the RBAC docs. Or an additional decorator to declare these objects on the handler functions.

rosscdh:
docs from readthe docs seem to be missing for both role and clusterrole

```
  - apiGroups: [apiextensions.k8s.io]
[...]

Issue #12, opened 2019-04-01
Documentation


nolar:
The project is now in the proof-of-concept stage, and lacks the documentation (beside the docstrings and examples).

It needs some normal hosted HTML docs (e.g. on ReadTheDocs).

* [x] Hosting configured.
* [x] Builds configured.

Specifically, some topics to not forget:

* [x] Typical patterns to solve typical problems (similar to examples). E.g., children object creation.
* [x] Deployment patterns: namespace isolation, deployment and RBAC-based service accounts 
 (minimum rules needed for the framework, not the general recommendations), etc. See #17.
* [x] Getting starting (or quick-start) guide (isn't it an example/01-minimal?).
* [x] The requirement for `**kwargs` or `**_` (for no-linting) for the forward-compatibility with the new keywords passed to the existing handlers by the new versions of the framework (part of the DSL).
* [x] The full list of all the available keywords now: `body`, `meta`, `spec`, `status`, `patch`, `logger`, `retry`, `diff`, `old`, `new`.
* [x] How the state is persistent in k8s itself, point out that the framework is stateless (same as the web servers).
* [x] What happens when the operator is killed during the handler, and how to avoid the duplicated side-effects (i.e. idempotent handlers).
* [x] What happens when the changes are applied when the operator is dead or fails permanently due to the bugs (the last-seen state preservation and comparison), i.e. that no changes are left behind unnoticed.
* [x] Strategies to isolate the operators to the namespaces; but still to have the cluster-wide operators (cluster CRDs or all-namespace monitoring, as e.g. Postgres operators, i.e. not already domain-specific ad-hoc solutions).
* [x] Alternatives: CoreOS Operator Framework, all Go-based solutions, etc). Differences and similarities, and the self-positioning of Kopf.

PR #50, opened 2019-04-29
Fix the docs for the RBAC examples, and add "get" permission on peering

nolar:
> Issue : #12, #49 

The docs were missing the cluster-vs-namespaced object separation.

And there were no "GET" permission on the peering object by its name. Now, it is added to address our incident (see #49 background).

These docs will be later affected by #36.


dneuhaeuser-zalando:
:+1:

PR #61, opened 2019-05-12
Test the routines for the progress status storage

nolar:
> Issue : #13 

Only the tests and few little edge-case fixes for the routines handling the progress storage.

The progress storage is used by the framework to keep the status of the handlers, their errors or delays, and the multi-handler processing in general.

The progress struct itself is already documented in the `kopf.structs.progress` docstring with a sample, see here: https://github.com/zalando-incubator/kopf/blob/master/kopf/structs/progress.py

### DIFF
--- a/tests/test_progress.py
+++ b/tests/test_progress.py
@@ -281,7 +281,7 @@ def test_store_failure(handler, expected, body):
 
 @pytest.mark.parametrize('result, body, expected', [
 
-    # With no result, it updates only tghe progress.
+    # With no result, it updates only the progress.
     (None,
      {},
      {'status': {'kopf': {'progress': {'some-id': {'stopped': TS0_ISO,
--- a/tests/test_progress.py
+++ b/tests/test_progress.py
@@ -1,0 +1,332 @@
+import copy
+import datetime
+from unittest.mock import Mock
+
+import freezegun
+import pytest
+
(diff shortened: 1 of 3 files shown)


tests/test_progress.py
```diff
@@ -0,0 +1,332 @@
...
+    origbody = copy.deepcopy(body)
+    patch = {}
+    store_failure(body=body, patch=patch, handler=handler, exc=Exception("some-error"))
+    assert patch == expected
+    assert body == origbody  # not modified
+
+
+@pytest.mark.parametrize('result, body, expected', [
+
+    # With no result, it updates only tghe progress.
```
samurang87: typo :)

nolar: @samurang87 Fixed.

samurang87:
There is an issue with a dependency. Do we need the testing requirements in requirements.txt at all?

nolar:
@samurang87 `freezegun` is Apache-2.0 licensed, this license is compatible with MIT: https://opensource.zalando.com/docs/resources/licensing/

And yes, the requirements are needed. Otherwise, the CI/CD tests will not run in Travis CI. (The `setup.py` does not include `requirements.txt`, it is vice versa: `requirements.txt` depends on `.` plus all the development/testing tools.).

PR #63, opened 2019-05-14
Test watching and queueing of the events into workers/handlers

nolar:
> Issue : #13 

The main goal is to add tests for the asyncio queueing multiplexer ("reactor") of Kopf.

It is incredibly difficult to test async logic without polluting the main codebase with numerous events/pause-points. 

Instead, the code is restructured to test the individual methods of this reactor: event streaming, event watching, and event queueing. This goes against the "best practices" of test design, as it now tests the specific implementation, not the behaviour from outside. But this was the only way to do this.

Some terminology hints to make the review easier to understand the reactor:

* **Streaming** means getting the events from the Kubernetes client's `Watch.stream()` synchronous method, and yielding them into the asynchronous reactor safely (i.e. non-blocking). It is consumed by the _watching_ task.
* **Watching** means getting these events from the infinite never-exiting stream, filtering out some of them silently (e.g. ERROR events cause the stream restart), and passing the normal expected events to the reactor's _queueing_ task.
* **Queueing** (the core of the "reactor") means getting the events to be handled in one stream, and then multiplexing (routing, distributing) them into the per-object (per-uid) queues, starting the workers tasks for each of these queues or reusing the existing ones, and garbage-collecting both the workers & the queues after some idle timeout of no events on the stream (~5s).
* **Batching** of the events means that only the last one of the events should be handled (see #42), while all other events in the small time-window (~0.1s) are ignored.
* **Handling** means getting the events of a specific object in its own dedicated asyncio task (thread), and calling the handler functions. — But this part is out of this PR, and will be done later.

It might be helpful to also read on the [Reactor pattern](https://en.wikipedia.org/wiki/Reactor_pattern).

As a side-effect of testing, few not-yet-detected issues were fixed:

* The watching logic was extracted #62 — no logic changes, just some code restructuring.
* The watcher & the workers now exit immediately — by using the EOS (end-of-stream) pseudo-event; previously, they exited after some timeout (5s), making the tests slow. The workers are given some timeout (2s) to exit gracefully before being force-killed, but this was not needed in the tests (might be necessary in the synchronous handlers in real operators).
* All hard-coded timing constants are now tweakable if needed.


nolar:
Rebased on the most fresh master.

PR #69, opened 2019-05-22
Refactor the FIELD pseudo-cause into the catch-all events and a field filter

nolar:
> Issue : #13 

The `FIELD` pseudo-cause was introduced as an easy way to mark the per-field handlers (`@kopf.on.field`), and then to select them on the event handling. The field handlers were handled separately, after all the object handlers.

Over time, this pseudo-cause was merged into the overall handling routine, and has become a problem: it prevents further refactoring (of event/cause detection and handling), which in turn is needed for the test coverage.

----

This PR is a set of changes extracted from a bigger branch for the ease of code review (one small part at a time). It does NOT change the behaviour of the framework or of any handlers; it just changes the internal implementation of it:

* No more `FIELD` pseudo-causes in the code.
* The catch-all handlers (`event==None`) with `field` property set are now used — i.e. any event as long as this field is in the diff. Efficiently, it means only the `UPDATE` events — same as before.

----

In addition, the tests of handler selection were improved to reflect all cases; and to make the test ids more readable.

### DIFF
--- a/tests/registries/test_handler_matching.py
+++ b/tests/registries/test_handler_matching.py
@@ -4,7 +4,6 @@ from unittest.mock import Mock
 import pytest
 
 from kopf import SimpleRegistry, GlobalRegistry
-from kopf.reactor.registries import FIELD
 
 
 # Used in the tests. Must be global-scoped, or its qualname will be affected.
@@ -12,13 +11,12 @@ def some_fn():
     pass
 
 
-@pytest.fixture(params=['simple', 'global'])
+@pytest.fixture(params=[
+    pytest.param(SimpleRegistry, id='in-simple-registry'),
+    pytest.param(GlobalRegistry, id='in-global-registry'),
+])
 def registry(request):
-    if request.param == 'simple':
-        return SimpleRegistry()
-    if request.param == 'global':
-        return GlobalRegistry()
-    raise Exception(f"Unsupported registry type: {request.param}")
+    return request.param()
 
 
 @pytest.fixture()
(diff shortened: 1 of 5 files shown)


tests/registries/test_handler_matching.py
```diff
@@ -30,72 +28,123 @@ def register_fn(registry, resource):
...
+    pytest.param(None, id='without-diff'),
+    pytest.param([], id='with-empty-diff'),
+])
+def cause_no_diff(request, resource):
+    return Mock(resource=resource, event='some-event', diff=request.param)
 
 
-@pytest.fixture()
+@pytest.fixture(params=[
+    pytest.param([('op', ('some-field',), 'old', 'new')], id='with-field-diff'),
```
MahshidZ: what does 'op' mean here?

tests/registries/test_handler_matching.py
```diff
@@ -30,72 +28,123 @@ def register_fn(registry, resource):
...
+    handlers = registry.get_handlers(cause_any_diff)
     assert handlers
 
 
-def test_relevant_event_handlers_found(registry, register_fn, cause_no_diff):
-    register_fn(some_fn, event='some-event')
+def test_catchall_handlers_with_field_found(cause_with_diff, registry, register_fn):
+    register_fn(some_fn, event=None, field='some-field')
+    handlers = registry.get_handlers(cause_with_diff)
+    assert handlers
```
MahshidZ: Do you assert when this is not empty? Does it mean it catches all the handlers?

tests/registries/test_handler_matching.py
nolar: Operation. Usually it is one of: add, change, remove. In this case, it is irrelevant, as the object-under-test is the cause & handlers, not the diff — but a list of 4-item tuples is expected in the internal for-cycle, no matter what's inside, so we give it.

tests/registries/test_handler_matching.py
nolar: Yes. It means, that something is there, i.e. non-empty. 

There is only one handler registered in the previous line, so technically it also mean that all handlers are returned (for that field: since both then handler and the cause contain this field — in this specific test).

PR #70, opened 2019-05-22
Fix the detection of field deletion in diffs

nolar:
> Issue : #13

As accidentally detected during #69 — this fixes a little issue with `KeyError` raised when the field is removed from the object, and a field-handler is called for that event, and fails to "resolve" the old/new value for the field-handler (i.e. a subset of a bigger diff-object), as the field is not in one of those per-object old/new structs.

### DIFF
--- a/kopf/structs/diffs.py
+++ b/kopf/structs/diffs.py
@@ -11,12 +11,20 @@ DiffPath = Tuple[str, ...]
 DiffItem = Tuple[DiffOp, DiffPath, Any, Any]
 Diff = Sequence[DiffItem]
 
+_UNSET = object()
 
-def resolve(d: Mapping, path: DiffPath):
-    result = d
-    for key in path:
-        result = result[key]
-    return result
+
+def resolve(d: Mapping, path: DiffPath, default=_UNSET):
+    try:
+        result = d
+        for key in path:
+            result = result[key]
+        return result
+    except KeyError:
+        if default is _UNSET:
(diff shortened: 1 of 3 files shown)


kopf/structs/diffs.py
```diff
@@ -11,12 +11,20 @@
 DiffItem = Tuple[DiffOp, DiffPath, Any, Any]
 Diff = Sequence[DiffItem]
 
-
-def resolve(d: Mapping, path: DiffPath):
-    result = d
-    for key in path:
-        result = result[key]
-    return result
+_UNSET = object()
```
samurang87: does the private method underscore for classes also work for global vars?

nolar: Python has no support for _protected_ methods in a classic OOP way. It is a human convention in Python — to underscore what is not supported to be used outside, and to never use anything starting with the underscore (though, technically, nothing will stop users from doing so).

Here, it is used for a marker of the function signatures, to distinguish from `None` as a legitimate value. This is a relatively common pattern for this case: e.g., in [configparser](https://github.com/python/cpython/blob/45d75faa7211592509c7c9da148a74687f182894/Lib/configparser.py#L354-L357) of stdlib.

PS: There is also some "magic" for _pseudo-private_ methods with "double-underscore mangling". It only works in the classes, as it prefixes the methods/fields with the class name. If there is no class, there is nothing to use as a prefix.

PR #71, opened 2019-05-24
Consolidate all K8s API calls in one place

nolar:
> Issue : #13

This is purely a refactoring: moving all Kubernetes-client-related routines into one place. 

And a lot of new tests for these K8s-client call (in addition to the existing ones).

Previously, the API calls were spread all over the codebase. And the official Kubernetes client is quite wordy, so the places were also unnecessary wordy (not to their main point).

----

**Few notes:**

The peering objects (cluster-scoped, namespace-scoped, and the legacy one) use exactly the same unified logic now, with no exceptions.

There are **no changes** in the APIs, DSLs, signatures, behaviour or anything user-facing or k8s-facing. Only the internal code moves. Except for this:

* When patching AND the object is a cluster-scoped resource (no namespace by design), `patch_cluster_custom_object()` is used instead of `patch_namespaced_custom_object(namespace=None)`, which would fail.  This would also affect the patching of peering objects (cluster-scoped and legacy), but is fixed in place.

Patching of a namespaced resource uses the namespace of the resource itself (even for the cluster-scoped operator).

----

This PR is also a part of the following **chain of changes**:

* _(This PR):_ Extraction of the K8s-API-related methods to a single place, both for served objects and for peering objects (the logic must be aligned).
  * _(Needed for):_ Testing of peering logic, regardless of the API calls. #13
  * _(Needed for):_ Testing of how the framework handles the events and which API methods it calls (mocking the central place instead of individual API client calls all around). #13 
    * _(Needed for):_ Safe changes of the handling logic (i.e. silent spies, etc). <<< **MAIN GOAL.**
  * _(Needed for):_ Switching to pykube-ng (#15), and generally being K8s-client-agnostic.
  * _(Needed for):_ Making the whole framework async, eliminating the blocking API calls (or putting them to the asyncio thread executors).


kopf/k8s/events.py
```diff
@@ -0,0 +1,52 @@
...
+        generate_name='kopf-event-',
+    )
+    body = kubernetes.client.V1beta1Event(
+        metadata=meta,
+
+        action='Action?',
+        type=type,
+        reason=reason,
+        note=message,
+        # message=message,
```
samurang87: are you leaving these comments intentionally? :)

nolar: Yes. 

I struggle to identify which fields do what here, and they are duplicating each other (but not all of them): e.g. `note` ~ `message`, `regarding` ~ `related`, `event_time` ~ `deprecated_first_timestamp`; also some are of unclear purpose, and are not shown anywhere, e.g. `action`.

I will get back to them later, probably when I will be switching Kopf to pykube-ng — just to see how it is better done with other clients.

This might also be needed for multi-version K8s compatibility, as these fields look like some transitional stage between Kubernetes versions, so it might happen that these lines should be uncommented later. — E.g., most likely, it will be a switch from `V1beta1Event` to something like `V1Event` class.

nolar: @samurang87 And immediately related: #81 — with the cleanup of these fields (on another occasion).

PR #73, opened 2019-05-25
Docs: use stable Workloads API (apps/v1)

hjacobs:
> Issue: #12 

Use the stable Kubernetes workloads API version, see https://kubernetes.io/docs/concepts/workloads/controllers/deployment/


nolar:
@hjacobs Thanks! 👍

Issue #76, opened 2019-05-27
issue running the first example


sebgoa:
I tried the first example and got:

```
sebair: 01-minimal (master)$ kopf run example.py --verbose
[2019-05-27 09:05:16,862] kopf.config          [DEBUG   ] configured via kubeconfig file
[2019-05-27 09:05:18,429] kopf.reactor.peering [WARNING ] Default peering object not found, falling back to the standalone mode.
[2019-05-27 09:05:38,900] kopf.reactor.handlin [DEBUG   ] [default/kopf-example-1] First appearance: {'apiVersion': 'zalando.org/v1', 'kind': 'KopfExample', 'metadata': {'annotations': {'kubectl.kubernetes.io/last-applied-configuration': '{"apiVersion":"zalando.org/v1","kind":"KopfExample","metadata":{"annotations":{},"labels":{"somelabel":"somevalue"},"name":"kopf-example-1","namespace":"default"},"spec":{"duration":"1m","field":"value","items":["item1","item2"]}}\n'}, 'creationTimestamp': '2019-05-27T07:05:35Z', 'generation': 1, 'labels': {'somelabel': 'somevalue'}, 'name': 'kopf-example-1', 'namespace': 'default', 'resourceVersion': '235142214', 'selfLink': '/apis/zalando.org/v1/namespaces/default/kopfexamples/kopf-example-1', 'uid': 'd359fea7-804d-11e9-b655-42010a80024c'}, 'spec': {'duration': '1m', 'field': 'value', 'items': ['item1', 'item2']}}
[2019-05-27 09:05:38,900] kopf.reactor.handlin [DEBUG   ] [default/kopf-example-1] Adding the finalizer, thus preventing the actual deletion.
[2019-05-27 09:05:38,900] kopf.reactor.handlin [DEBUG   ] [default/kopf-example-1] Patching with: {'metadata': {'finalizers': ['KopfFinalizerMarker']}}
[...]

PR #81, opened 2019-05-28
Use the core v1 API instead of v1beta1, for GKE>=1.12 compatibility

nolar:
> Issue : #76 

Starting from GKE 1.12 (Google Kubernetes Engine), the legacy APIs are disabled by default — `disable-legacy-endpoints=true` is added on the cluster creation. [Official doc](https://cloud.google.com/kubernetes-engine/docs/how-to/protecting-cluster-metadata#disable-legacy-apis).

As such, events of `v1beta1` are not accessible, and Kopf fails on the simple examples (see #76). `v1` events should be used instead.

This PR switches to the `v1` events. And v1beta1 are now disabled for Minikube, so it will fails if accidentally used. Also, it adds the unit-tests that were absent since #71 .

Beside the tests, I have manually tested the Kopf example operator to run in GKE — now, it works (before the fix, it was failing on the event posting).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ matrix:
     - python: "3.8-dev"
 
 before_script:
-  - scripts/minikube-for-travis.sh
+  - tools/minikube-for-travis.sh
 
 script:
   - pytest -v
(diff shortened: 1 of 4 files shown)


smileisak:
I was working on it. I did not see the issue 👍. Nice work !

nolar:
@smileisak Oops. Sorry. 

Since contributors started to join, I think, we should start syncing the allocation of issues by assigning them. I've self-assigned those I work on right now. Feel free to choose a free one: https://github.com/zalando-incubator/kopf/issues (or create a new one if the bug/enhancement is not mentioned).

PR #83, opened 2019-05-28
Document the operator maintenance

### LINKED ISSUE #12: Documentation

### LINKED ISSUE #75: Add warning to docs about finalizers.
FYI. You may want to add a warning to the documentation about problems that can occur due to the way in which finalizers are used on a custom CRD. In particular, if the operator is deployed to a namespace and setup in standalone mode to monitor just the same namespace, if the operator is deleted before the custom CRD objects, then when the namespace is deleted, it will get stuck in terminating state.

This is especially problematic, because deletion of a project doesn't really guarantee any order of deletion where owner references and blocking conditions haven't been specified. Thus deletion of a namespace can result in the operator being deleted first, without there being a need to have explicitly delete the operator first to cause this issue. Even if the operator is deployed in a separate project, same issue can arise were the operator were separately deleted.

The only solution to this is to identify what the custom CRD object is that is blocking deletion of the namespace, and manually edit it to remove the finalizer. I don't know of any better solution at the moment, or how you could rework kopf to not be affected by the issue.

nolar:
> Issue : closes #75, closes #12

Add some pages for the maintenance of the Kopf-based operators, describing their behaviour on downtime, on restarts, how they keep the persistence.

Add the troubleshooting page for some unexpected behaviour, to make it clear (e.g. #75).

_A new section was needed, as this kind of information does not fit to Tutorial, Features, About Kopf, and not even to the Developers Manual (which is about developing Kopf itself, not the Kopf-based operators)._

Preview:

* https://kopf-nolars-fork.readthedocs.io/en/docs-maintenance/continuity.html
* https://kopf-nolars-fork.readthedocs.io/en/docs-maintenance/idempotence.html
* https://kopf-nolars-fork.readthedocs.io/en/docs-maintenance/troubleshooting.html

### DIFF
--- a/docs/continuity.rst
+++ b/docs/continuity.rst
@@ -1,0 +1,58 @@
+==========
+Continuity
+==========
+
+Persistence
+===========
+
+Kopf does not have any database. It stores all the information directly
+on the objects in the Kubernetes cluster (which means `etcd` usually).
+All information is retrieved and stored via the Kubernetes API.
+
+Specifically:
+
+* The cross-operator exchange is performed via the peering objects.
+  See :doc:`peering` for more info.
+* The last handled state of the object is stored in ``metadata.annotations``.
+  It is used for the diff calculation on the changes.
+* The handler status (failures, successes, retries, delays) is stored
+  in ``status.kopf.progress`` (with ``status.kopf`` reserved for any
+  framework-related information in the future).
+
+
+Restarts
+========
+
+It is safe to kill the operator's pod (or process), and allow it to restart.
+
+The handlers that succeeded previously, will not be re-executed.
+The handlers that did not execute yet, or were scheduled for retrying,
+will be retried by a new operators pod/process from the point where
+the old pod/process was terminated.
+
+The restart of an operator will only affect the handlers currently being
+executed in that operator at the moment of termination, as there is
+no record that they have succeeded.
+
+
+Downtime
+========
+
+If the operator is down and not running, any changes to the objects
+are ignored and not handled. They will be handled when the operator starts:
+every time a Kopf-based operator starts, it lists all objects of the served
+resource kind, and checks for their state; if the state has changed since
+the object was last handled (no matter how long time ago),
+a new handling cycle starts.
+
+Only the last state is taken into account. All the intermenient changes
+are accumulated and handled together.
+This corresponds to the Kubernetes'es concept of eventual consistency
+and level triggering (as opposed to edge triggering).
+
+.. warning::
+    If the operator is down, the objects cannot be deleted,
+    as they contain the Kopf's finalizers in ``metadata.finalizers``,
+    and Kubernetes blocks the deletion until all finalizers are removed.
+    If the operator is not running, the finalizers will never be removed.
+    See: :ref:`finalizers-blocking-deletion` for a work-around.
--- a/docs/continuity.rst
+++ b/docs/continuity.rst
@@ -14,7 +14,7 @@ Specifically:
 * The cross-operator exchange is performed via the peering objects.
   See :doc:`peering` for more info.
 * The last handled state of the object is stored in ``metadata.annotations``.
-  It is used for the diff calculation on the changes.
+  It is used to calculate diffs upon changes.
 * The handler status (failures, successes, retries, delays) is stored
   in ``status.kopf.progress`` (with ``status.kopf`` reserved for any
   framework-related information in the future).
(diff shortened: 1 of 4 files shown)


docs/continuity.rst
```diff
@@ -0,0 +1,58 @@
...
+Kopf does not have any database. It stores all the information directly
+on the objects in the Kubernetes cluster (which means `etcd` usually).
+All information is retrieved and stored via the Kubernetes API.
+
+Specifically:
+
+* The cross-operator exchange is performed via the peering objects.
+  See :doc:`peering` for more info.
+* The last handled state of the object is stored in ``metadata.annotations``.
+  It is used for the diff calculation on the changes.
```
samurang87: "it is used to calculate diffs upon changes" ... I guess

docs/continuity.rst
```diff
@@ -0,0 +1,58 @@
...
+  in ``status.kopf.progress`` (with ``status.kopf`` reserved for any
+  framework-related information in the future).
+
+
+Restarts
+========
+
+It is safe to kill the operator's pod (or process), and allow it to restart.
+
+The handlers that succeeded previously, will not be re-executed.
```
samurang87: remove the comma :)

docs/continuity.rst
```diff
@@ -0,0 +1,58 @@
...
+The handlers that did not execute yet, or were scheduled for retrying,
+will be retried by a new operators pod/process from the point where
+the old pod/process was terminated.
+
+The restart of an operator will only affect the handlers currently being
```
samurang87: Restarting an operator...

docs/continuity.rst
```diff
@@ -0,0 +1,58 @@
...
+========
+
+If the operator is down and not running, any changes to the objects
+are ignored and not handled. They will be handled when the operator starts:
+every time a Kopf-based operator starts, it lists all objects of the served
+resource kind, and checks for their state; if the state has changed since
+the object was last handled (no matter how long time ago),
+a new handling cycle starts.
+
+Only the last state is taken into account. All the intermenient changes
```
samurang87: I guess you have a typo here (intermediate? intervening?)

docs/continuity.rst
```diff
@@ -0,0 +1,58 @@
...
+are accumulated and handled together.
+This corresponds to the Kubernetes'es concept of eventual consistency
```
samurang87: This corresponds to Kubernetes's (🤔 ) concept of ...

docs/idempotence.rst
```diff
@@ -0,0 +1,55 @@
+===========
+Idempotence
+===========
+
+Kopf provides the tools to make the handlers idempotent.
```
samurang87: "Kopf provides tools..."

nolar:
@samurang87 I've fixed some grammar issues :-)

PR #86, opened 2019-05-29
Add the silent spy-handlers for the raw events, no persistence

### LINKED ISSUE #30: Silent handlers (spies)

nolar:
> Issue: closes #30

Add the simplistic event-handlers for spying on the low-level events of other resources,
not belonging to the operator (and not storing any status fields or finalizers or so).

The syntax is:

```python
import kopf

@kopf.on.event('zalando.org', 'v1', 'kopfexamples')
def event_fn(event, **kwargs):
    pass
```

The difference from the regular handlers is documented in the `docs/handlers.rst` (see the changed files). Briefly:

* No state storage on the objects.
* No annotations or finalizers added to the objects.
* No k8s-events posted for the activity (only the stdout/stderr logs).

Altogether, this means "silent" — the objects do not know they are being watched & handled (i.e. spied).

Checklist:

* [ ] Tests.
* [ ] Documentation.

### DIFF
--- a/docs/handlers.rst
+++ b/docs/handlers.rst
@@ -47,7 +47,7 @@ To register a handler for an event, use the `kopf.on` decorator::
     def my_handler(spec, **_):
         pass
 
-All available decorators is described below.
+All available decorators are described below.
 
 
 Arguments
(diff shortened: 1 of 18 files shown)


docs/handlers.rst
```diff
@@ -60,22 +60,60 @@ The following keyword arguments are available to the handlers
...
 
 ``**kwargs`` (or ``**_`` to stop the linting warnings on the unused variables)
 is required for the forward compatibility: the framework can add new keywords
 in the future, and the existing handlers should accept them and not break.
 
 
+Event handlers
+==============
+
+The low-level events can be intercepted and handled silently, without
```
samurang87: Low-level events...

examples/08-events/README.md
```diff
@@ -0,0 +1,52 @@
+# Kopf example with spy-handlers for the raw events
+
+Kopf stores its handler status on the objects' status field.
+This can be not desired when the objects do not belong to this operator,
+but a probably served by some other operator, and are just watched
+by the current operator, e.g. for their status fields.
+
+The event-handlers can be used as the silent spies on the raw events:
```
samurang87: Event-handlers can be used...

docs/handlers.rst
```diff
@@ -60,22 +60,60 @@ The following keyword arguments are available to the handlers
 * ``spec`` as an alias for ``body['spec']``.
 * ``meta`` as an alias for ``body['metadata']``.
 * ``status`` as an alias for ``body['status']``.
+* ``namespace``, ``name``, ``uid`` to identify the object being handled,
+  and are aliases for the respective fields in ``body['metadata']``.
+* ``logger`` is a per-object logger, with the messages prefixed with the object's namespace/name.
+* ``patch`` is a dict with the object changes to be applied after the handler.
+
+The high-level cause-handlers have additional keyword arguments reflecting their status:
```
samurang87: High-level cause-handlers... (and later, low-level event-handlers...)

examples/08-events/example.py
```diff
@@ -0,0 +1,14 @@
+import kopf
+
+# Marks for the e2e tests (see tests/e2e/test_examples.py):
+E2E_TRACEBACKS = True
+
+
+@kopf.on.event('zalando.org', 'v1', 'kopfexamples')
+def event_fn_with_error(**kwargs):
+    raise Exception("Oops!")
```
samurang87: lol

kopf/reactor/registries.py
```diff
@@ -108,29 +123,56 @@ class GlobalRegistry(BaseRegistry):
...
         """
         resource = Resource(group, version, plural)
-        registry = self._handlers.setdefault(resource, SimpleRegistry())
+        registry = self._cause_handlers.setdefault(resource, SimpleRegistry())
         registry.register(event=event, field=field, fn=fn, id=id, timeout=timeout)
         return fn  # to be usable as a decorator too.
 
+    def register_event_handler(self, group, version, plural, fn, id=None):
+        """
+        Register an additional handler function for the low-level events.
```
samurang87: for low-level events

Issue #87, opened 2019-05-30
Limit size of exception details in message field of event.


GrahamDumpleton:
When an exception occurs, the full exception traceback is being added to the `message` field of an event. When this is posted to Kube, it result in an error as it can be too big to store.

```
HTTP response headers: HTTPHeaderDict({'Cache-Control': 'no-store', 'Content-Type': 'application/json', 'Dat
e': 'Thu, 30 May 2019 02:05:15 GMT', 'Content-Length': '392'})
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Event \"k
opf-event-5sqgc\" is invalid: message: Invalid value: \"\": can have at most 1024 characters","reason":"Inva
lid","details":{"name":"kopf-event-5sqgc","kind":"Event","causes":[{"reason":"FieldValueInvalid","message":"
Invalid value: \"\": can have at most 1024 characters","field":"message"}]},"code":422}
```

Should perhaps restrict what is used to lowest stack frame and exception description.

nolar:
@GrahamDumpleton Can you please tell which exception is this? Or share a code snippet to reproduce it. 

Technically, a string representation of an exception is used (`f"{exc}"` in `kopf.structs.status. store_failure()`), which is usually just its type+message, not the stacktrace.

nolar:
Fixed in `kopf==0.14`: https://github.com/zalando-incubator/kopf/releases/tag/0.14

Specifically:

* The event messages are cut to 1024 chars, keeping the start & end of the original string.
* Any errors in event posting are logged, but do not break the operator anymore (as events are auxiliary by their nature).

PR #88, opened 2019-05-30
Suppress errors in event posting, and prevent them when possible

nolar:
> Issue : #87 

Handle the case when Kopf fails to post a Kubernetes event (for logging), and by doing so interrupts the handling cycle. The events are auxiliary by their nature, not mandatory. Their errors should not break the framework's or operator's behaviour. Yet they must be noticed (e.g. in logs).

With this PR, the event posting is changed:

* If any API error happens during posting an event, a warning is logged, but the operator continues to handle the resources. The event info and the error are both put to the logs for later investigation.
* The messages are cut to 1024 characters, keeping the starting and ending parts, and putting an infix in the middle. At least something will be posted, which is better than nothing at all.

By occasion also move the shared parts to conftest, as now we have more than 1 test for events.

### DIFF
--- a/kopf/k8s/events.py
+++ b/kopf/k8s/events.py
@@ -1,9 +1,12 @@
 import datetime
 import logging
 
-import kubernetes
+import kubernetes.client.rest
 
 logger = logging.getLogger(__name__)
+
+MAX_MESSAGE_LENGTH = 1024
+CUT_MESSAGE_INFIX = '...'
 
 
 def post_event(*, obj, type, reason, message=''):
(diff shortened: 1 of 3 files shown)


kopf/k8s/events.py
```diff
@@ -11,6 +17,13 @@ def post_event(*, obj, type, reason, message=''):
     now = datetime.datetime.utcnow()
     namespace = obj['metadata']['namespace']
 
+    # Prevent a common case of event posting errors but shortening the message.
+    if len(message) > MAX_MESSAGE_LENGTH:
+        infix = CUT_MESSAGE_INFIX
```
samurang87: infix? 🤔

samurang87: much grammar wow

Issue #92, opened 2019-06-01
kopf disallow importing modules on path


trondhindenes:
Given the following app structure:
```
╰─ cat app.py 
import kopf
import json

from MyMod.lib1 import Lib1
l = Lib1

@kopf.on.create('testing.com', 'v1', 'mycrd')
def create_fn(body, **kwargs):
    print(f'A handler is called with body: {json.dumps(body, indent=4, sort_keys=True)}')
```
```
╰─ ls MyMod 
__init__.py  lib1.py
```

```
╰─ cat MyMod/lib1.py 
class Lib1:
    def __init__():
        pass
    def doit(what):
        pass
```

In other words, in the app dir, I have a regular python module

If I execute the app.py file with regular Python, it just exits (as is expected), without throwing any error, meaning that all the imports work as expected.

However, if I run `pipenv run kopf run app.py`, it throws `ModuleNotFoundError: No module named 'MyMod'`.

## Expected Behavior
apps run using kopf should be able to import local modules just fine

## Actual Behavior
kopf throws (see above

## Steps to Reproduce the Problem
in a blank dir:
```
pip install pipenv
PIPENV_VENV_IN_PROJECT=true pipenv install --dev --python python3.7
```
create files as described above
`pipenv run kopf run app.py`
## Specifications

  - Platform:
  - Kubernetes version: *(use `kubectl version`)*
  - Python version: *(use `python --version`)*
  - Python packages installed: *(use `pip freeze --all`)*

```
```

trondhindenes:
as far as I can see, all that's missing is adding `curdir`to sys.path prior to loading.

nolar:
@trondhindenes Thanks for detailed report.

Kopf uses the standard Python's importing system and does not hack the Python's importing mechanics, as it can affect different projects with different layouts in differently ways — and this importing logic falls out of scope of a Kubernetes operator framework. 

Putting the current script's dir (but not the current work dir!) to `sys.path` is one of the Python's default tricks, as officially documented. Since the current script is `kopf` from the virtualenv's `bin` directory, this is not much helpful.

The easiest way you can try to use is to hack the importing paths with [`PYTHONPATH`](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPATH) env var:

```bash
PYTHONPATH=/home/$USER/src/myproj/  pipenv run kopf run ...
# OR:
PYTHONPATH=.  pipenv run kopf run ...
```

[...]

PR #93, opened 2019-06-01
Add file's dir to sys.path

trondhindenes:
# One-line summary
Add current dir to sys.path prior to imports

> Issue : #92 

## Description
Fixes the bug where modules weren't importable from current path since `current dir` was missing from `sys.path` (where modules are resolved from)

## Types of Changes

### DIFF
--- a/kopf/reactor/loading.py
+++ b/kopf/reactor/loading.py
@@ -14,9 +14,11 @@ Currently, two loading modes are supported, both are equivalent to Python CLI:
 Multiple files/modules can be specified. They will be loaded in the order.
 """
 
+import sys
 import importlib
 import importlib.util
 import os.path
+
 
 
 def preload(paths, modules):
@@ -25,6 +27,8 @@ def preload(paths, modules):
     """
 
     for path in paths:
+        print(path  )
+        sys.path.append(os.path.abspath(os.path.join(path, os.pardir)))
         name, _ = os.path.splitext(os.path.basename(path))
         spec = importlib.util.spec_from_file_location(name, path)
         module = importlib.util.module_from_spec(spec)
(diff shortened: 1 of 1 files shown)


kopf/reactor/loading.py
```diff
@@ -14,17 +14,20 @@
 Multiple files/modules can be specified. They will be loaded in the order.
 """
 
+import sys
 import importlib
 import importlib.util
 import os.path
 
 
+
```
nolar: Python's conventions says it should be 2 empty lines before top-level functions. Can you please fix it here?

kopf/reactor/loading.py
```diff
@@ -14,17 +14,20 @@
...
 def preload(paths, modules):
     """
     Ensure the handlers are registered by loading/importing the files/modules.
     """
 
     for path in paths:
+        sys.path.append(os.path.abspath(os.path.curdir))
```
nolar: Since Kopf promises to mimic Python here, it is better to prepend the path here instead of appending.

Also, Python adds not the current directory, but the directory of the script being executed. Can be tested this way:

```bash
$ mkdir dir1
$ echo 'import sys; print(repr(sys.path))' > dir1/file1.py
$ python dir1/file1.py
['/Users/svasilyev/dir1', .....]
```

```suggestion
        sys.path.insert(0, os.path.abspath(os.path.dirname(path)))
```

Kopf has slightly enhanced semantics here with multiple scripts in one command line: `kopf run dir1/script1.py dir2/script2.py ...`. The dirs will be prepended in reverse order. But I don't think it is a big issue for now (can be fixed later).

The resulting `sys.path` will be `['.../proj/dir2', '.../proj/dir1', '/other/system/dirs']`. The relative modules/packages will be importable then.

nolar:
Sorry, I could edit the files, but couldn't resolve the merge conflicts in "not-my" PR, so had to make it separately in #133 — the original commits were preserve though.

trondhindenes:
sorry, I was on vacation. Thanks for taking this over @nolar

Issue #60, opened 2019-05-10
Handler for starting the per-object background threads/coroutines


nolar:
In some case, it is desired to have a background thread or async coroutine for every object existing in the cluster (if the operator developer decides so). 

Example use-case: polling the external services & APIs when the custom resource represents the jobs of an external system with its `status` field.

It can be done from the `@kopf.on.create()` handler. However, when the operator restarts, and there are no events on the object, no handler will ever be called, meaning the operator will not react to the external state changes.

With this handler, the problem can be solved:

```
import kopf

@kopf.on.seen('zalando.org', 'v1', 'kopfexamples')
def seen_fn(meta, **_):
    thread = threading.Thread(target=monitoring_fn, args=(meta['name'],))
    thread.start()
```

Other names: `@kopf.on.regain()`, `@kopf.on.restore()`.

----

**UNDECIDED:**

**Option A:**

The handler must be called only in the case when the operator restarts, gets a fresh list of the objects as part of the watch request, and notices there are some object already handled.

It must NOT be called when:

* When the object is actually created while the operator is running ("first seen" or "create" causes).
* When the operator starts and gets a fresh list of the objects as part of the watch request, and the objects are new and the `@kopf.on.create` handler should be called.

**Options B:**

* When the object is fully handled (including the creation), and now needs the monitoring — both in cases of creation, and the operator restart.

----

PR #96, opened 2019-06-02
Add "resuming" handlers for resources

nolar:
> Issue : #60 

## Description

Add the handlers for "resuming" (recalling? remembering?) an object in the operator:

```python
@kopf.on.resume('zalando.org', 'v1', 'kopfexamples')
async def on_resume(**kwargs):
    loop = asyncio.get_running_loop()
    loop.create_task(a_long_running_coroutine(**kwargs))

    thread = threading.Thread(target=a_long_running_function, kwargs=kwargs)
    thread.start()
```

This complements the existing handlers for creation/update/deletion cases, and is triggered when the object is neither created, nor updated, nor deleted, but the operator has started/restarted, and needs to be aware of that object somehow.

_**As a side-note:** I could not decide between "resume" and "recall" for proper naming. It is "resume", as it resumes some thread that was previously running in a now-dead operator. It is "recall", as it kind of "recalls" (remembers) the object: "ah, there is also such a thing with uid 123, which is not changed!". The naming can be changed before merged._

## Types of Changes

- Refactor/improvements
- Documentation / non-code

## Review

- [ ] Tests
- [ ] Documentation


nolar:
The first experience using it:

If I write

```
@kopf.on.resume('zalando.org', 'v1', 'kopfexamples')
def monitor_something(*, namespace, name, spec, logger, **kwargs):
    threading.Thread(...).start()
```

I expect that the thread is started either when a new object is created and the operator is running for some time, or when the object exists for some time and the operator has just restarted (i.e. either one is fresh new).

However, if the object was changed (e.g. its status) before the crash or restart, the `UPDATE` cause is detected, and the `@kopf.on.update` handlers are called. The `@kopf.on.resume` handlers are NOT called in case the object was changed — this **breaks the intuitive expectations and semantics of a background thread starter** (the original purpose of this handler).

Starting the threads from the `UPDATE`/`DELETE` handlers is not desired, as it means we have to keep track of the thread pool state per object — extra hassle.

The `RESUME` cause has to be handled even if the object has changed (but maybe not if deleted) — before or after the `UPDATE` handler in the same handling cycle.

PS: Or maybe there is no `RESUME` cause at all, and the `@kopf.on.resume` handlers are more like the event-handlers (`@kopf.on.event`), just happening on the initial object listing (maybe: and on creation), and not happening later.

nolar:
Replaced by an alternative approach in #105, which feels more intuitive.

PR #105, opened 2019-06-06
Add "resuming" handlers for resources (as co-handlers)

### LINKED ISSUE #60: Handler for starting the per-object background threads/coroutines

nolar:
> Issue : closes #60, replaces #96 

## Description

Add the handlers for "resuming" (recalling? remembering?) an object in the operator:

```python
@kopf.on.resume('zalando.org', 'v1', 'kopfexamples')
def on_resume(**kwargs): pass
```

This complements the existing handlers for creation/update/deletion cases, and is injected into the handling cycle when one of the mentioned cases is detected (and handled normally), but strictly when the operator has started/restarted, and found that the object existed beforehand.

Which implies that if the object was created during the operator downtime, both `@kopf.on.create` **AND** `@kopf.on.resume` handlers will be called. Similarly, of the object was changed during the operator downtime, both `@kopf.on.update` **AND** `@kopf.on.resume` will be called (and both will get a diff of the object).

---

_BUT: See #103 for the de-duplication of the same function if it is registered for few causes._

This "double-cause" patten becomes possible to start a background job for permanent object monitoring (the function will be called only once when the object appears: either while the operator is running and it is created, or when the operator starts and a new unhandled object is found, or when the operator starts and an existing handled object is found):

```python
@kopf.on.resume('zalando.org', 'v1', 'kopfexamples')
async def start_monotoring(**kwargs): pass
    loop = asyncio.get_running_loop()
    loop.create_task(a_long_running_coroutine(**kwargs))

    thread = threading.Thread(target=a_long_running_function, kwargs=kwargs)
    thread.start()
```

---

**Difference from #96:** Unlike a solution in #96, where the resuming handler is triggered when the object is neither created, nor updated, nor deleted, but noticed to exist, in this PR, the resume handler is called event if update/creation/deletion detected, as part of the relevant handling cycle (i.e. as one of the handlers on the list).

---

_**As a side-note:** I could not decide between "resume" and "recall" for proper naming. It is "resume", as it resumes some thread that was previously running in a now-dead operator. It is "recall", as it kind of "recalls" (remembers) the object: "ah, there is also such a thing with uid 123, which is not changed!". The naming can be changed before merged._

## Types of Changes

- Refactor/improvements
- Documentation / non-code

## Review

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/kopf/reactor/registries.py
+++ b/kopf/reactor/registries.py
@@ -59,7 +59,7 @@ class BaseRegistry(metaclass=abc.ABCMeta):
             @kopf.on.resume(...)
             def fn(**kwargs): pass
 
-        In normal case, the function will be called either on resource creation,
+        In normal cases, the function will be called either on resource creation
         or on operator restart for the pre-existing (already handled) resources.
         When a resource is created during the operator downtime, it is
         both creation and resuming at the same time: the object is new (not yet
@@ -115,7 +115,7 @@ class SimpleRegistry(BaseRegistry):
         for handler in self._handlers:
             if handler.event is None or handler.event == cause.event:
                 if handler.initial and not cause.initial:
-                    pass  # ignore initial handlers in the non-initial causes.
+                    pass  # ignore initial handlers in non-initial causes.
                 elif handler.field:
                     if any(field[:len(handler.field)] == handler.field for field in fields):
                         yield handler
--- a/kopf/reactor/registries.py
+++ b/kopf/reactor/registries.py
@@ -114,7 +114,9 @@ class SimpleRegistry(BaseRegistry):
         fields = {field for _, field, _, _ in cause.diff or []}
         for handler in self._handlers:
             if handler.event is None or handler.event == cause.event:
-                if handler.field:
+                if handler.initial and not cause.initial:
+                    pass  # ignore initial handlers in the non-initial causes.
+                elif handler.field:
                     if any(field[:len(handler.field)] == handler.field for field in fields):
                         yield handler
                 else:
--- a/kopf/reactor/registries.py
+++ b/kopf/reactor/registries.py
@@ -31,18 +31,48 @@ class BaseRegistry(metaclass=abc.ABCMeta):
     """
 
     def get_cause_handlers(self, cause):
-        return list(self.iter_cause_handlers(cause=cause))
+        return list(self._deduplicated(self.iter_cause_handlers(cause=cause)))
 
     @abc.abstractmethod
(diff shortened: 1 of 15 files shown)


kopf/reactor/registries.py
```diff
@@ -31,19 +31,49 @@ class BaseRegistry(metaclass=abc.ABCMeta):
...
+        single event/cause, even if it is registered with multiple decorators
+        (e.g. different filtering criteria or different but same-effect causes).
+
+        One of the ways how this could happen::
+
+            @kopf.on.create(...)
+            @kopf.on.resume(...)
+            def fn(**kwargs): pass
+
+        In normal case, the function will be called either on resource creation,
```
samurang87: normal cases

kopf/reactor/registries.py
```diff
@@ -84,7 +114,9 @@ def iter_cause_handlers(self, cause):
         fields = {field for _, field, _, _ in cause.diff or []}
         for handler in self._handlers:
             if handler.event is None or handler.event == cause.event:
-                if handler.field:
+                if handler.initial and not cause.initial:
+                    pass  # ignore initial handlers in the non-initial causes.
```
samurang87: in non-initial causes

samurang87:
I prefer "resume" :)

Issue #107, opened 2019-06-12
Label the sub-templates in the container resources (e.g. pods in jobs)


nolar:
## Actual Behavior

When I create a job, and label it with some parent-referencing labels, I expect the labels to be also applied to the pod templates:

```
kopf.label(job, {'parent-name': name})
```

The pod template lacks the labels, only the parent job objects gets it.

One way to work around this is to explicitly label the job and its template:

```
kopf.label(job, {'parent-name': name})
kopf.label(job['spec']['template'], {'parent-name': name})
```

## Expected Behavior

However, it would be better to do in one call:

```
# Explicitly:
kopf.label(job, {'parent-name': name}, nested=['spec.template'])

# Auto-guessing:
kopf.label(job, {'parent-name': name}, recursive=True)
```

PR #108, opened 2019-06-12
Label sub-templates recursively

nolar:
> Issue : #107, indirectly #13 

Now, it is a bit easier to label both the created object (e.g. jobs) and its sub-templates (e.g. pods in jobs):

```
kopf.label(job, {'parent-name': name}, nested=['spec.template'])
```

Also, few tests added on the utility functions used internally, just to be sure they work as expected.

The auto-guessing on which sub-fields are templates, which are not (as suggested in #107) is skipped for now — that might require enumerating the kinds&fields from the docs.

## Types of Changes


kopf/structs/dicts.py
```diff
@@ -0,0 +1,70 @@
...
+    elif isinstance(objs, collections.abc.Mapping):
+        yield objs
+        for subfield in (nested if nested is not None else []):
+            try:
+                yield resolve(objs, parse_field(subfield))
+            except KeyError:
+                pass
+    elif isinstance(objs, collections.abc.Iterable):
+        for obj in objs:
+            yield from walk(obj, nested=nested)
```
samurang87: must be the first time I see recursion in "real" code!

PR #109, opened 2019-06-12
Some concurrency fixes, some configurability

dbazhal:
# Trying to fix blocking operations, such as objects patching, event posting, etc

> Issue : #111 

## Description
1. Added config - kopf.config.WorkersConfig - class attributes based options to control async workers. So user can just import it before all other kopf modules and change values to required. Maybe not the best decision, but any.
2. So now can be configured number of next types of workers: queue processing, object patching, event creation.
3. Blocking synchronous operations such as event creation, changed objects patching, are now asynchronous (i guess).
4. Added kopf.config.EventsConfig.events_loglevel  to control what types of logs should trigger event sending - INFO/WARNING/ERROR/CRITICAL
    
Existing tests updated for new async functions.
E2E tests required to test performance(how fast large number of new/changed objects can be processed). Such tests would prove efficiency of kopf concurrency.

## Types of Changes

- Configuration change
- Refactor/improvements

## Tasks
_List of tasks you will do to complete the PR_
  - [x] Documentation
  - [ ] E2E performance tests
  - [ ] Review?

## Review
- [ ] Tests (don't know if there could be tests to do)
- [ ] Documentation (maybe?)


kopf/k8s/events.py
```diff
@@ -56,11 +64,15 @@ def post_event(*, obj, type, reason, message=''):
         event_time=now.isoformat() + 'Z',  # '2019-01-28T18:25:03.000000Z'
     )
 
+    api = kubernetes.client.CoreV1Api()
+    loop = asyncio.get_event_loop()
+    if not loop:
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
```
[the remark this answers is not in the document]
nolar: Also, since `post_event` is an async coroutine already, it runs in some event-loop. Therefore, `asyncio.get_running_loop()` would be sufficient and guaranteed to return an event-loop. And, [according to the docs](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop), it is recommended over `asyncio.get_event_loop()` because it is slightly faster/easier.

kopf/k8s/patching.py
```diff
@@ -20,20 +28,20 @@ def patch_obj(*, resource, patch, namespace=None, name=None, body=None):
...
+        'body': patch
+    }
+    patch_func = api.patch_cluster_custom_object
+    if namespace is not None:
+        request_kwargs['namespace'] = namespace
+        patch_func = api.patch_namespaced_custom_object
+    loop = asyncio.get_event_loop()
+    if not loop:
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
```
nolar: Same story with the excplicit loop management as in `post_event()`

kopf/reactor/queueing.py
```diff
@@ -247,8 +239,11 @@ def run(
     # The errors in the cancellation stage will be ignored anyway (never re-raised below).
     for task in pending:
         task.cancel()
-    cancelled, pending = loop.run_until_complete(asyncio.wait(pending, return_when=asyncio.ALL_COMPLETED))
-    assert not pending  # must be empty by now, the tasks are either done or cancelled.
+    if pending:
+        cancelled, pending = loop.run_until_complete(asyncio.wait(pending, return_when=asyncio.ALL_COMPLETED))
+        assert not pending  # must be empty by now, the tasks are either done or cancelled.
+    else:
+        cancelled = []
```
nolar: Can you please clarify what is happening here? 

I know that assertions in the main code are not good. The original idea was that it is impossible that anything is pending at this point. A regular exception can be thrown.

Do I get it right that it is because `asyncio.wait()` on an empty list fails? Still, I didn't get the part with `cancelled = []` — they still have to be `.result()`'ed few lines below.

dbazhal: Without this check :243 loop.run_until_complete will fail with
```
ValueError: Set of coroutines/Futures is empty.
```
(no pending tasks left). So this is ugly fix for that case.

kopf/k8s/events.py
```diff
@@ -1,15 +1,23 @@
+import asyncio
+import concurrent.futures
 import datetime
+import functools
 import logging
 
 import kubernetes.client.rest
 
+from kopf.config import WorkersConfig
```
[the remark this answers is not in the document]
dbazhal: done

kopf/config.py
```diff
@@ -11,6 +11,12 @@
 format = '[%(asctime)s] %(name)-20.20s [%(levelname)-8.8s] %(message)s'
 
 
+LOGLEVEL_INFO = logging.INFO
+LOGLEVEL_WARNING = logging.WARNING
+LOGLEVEL_ERROR = logging.ERROR
+LOGLEVEL_CRITICAL = logging.CRITICAL
```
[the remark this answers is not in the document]
dbazhal: I think it is better to help users to get used to work with kopf itself. Loglevels from logging module oblige some logic based on them - logging logic, so there is some solid concept of loglevels, and how they should be used. And here in kopf i suppose there will be some different logic around this levels, with not all of the logging loglevels included.
While writing this I understand now that it is not ok to duplicate logging loglevels, it's better to define them by themselves.

kopf/events.py
```diff
@@ -11,36 +11,62 @@
...
         _, exc, _ = sys.exc_info()
     reason = reason if reason else type(exc).__name__
     message = f'{message} {exc}' if message else f'{exc}'
-    return event(obj, reason=reason, message=message, type='Error')
+    await event_async(obj, reason=reason, message=message, type='Error')
+
+
+# Next 4 funcs are just synchronous interface for async event functions.
+def event(obj, *, type, reason, message=''):
+    asyncio.wait_for(event_async(obj, type=type, reason=reason, message=message), timeout=2)
```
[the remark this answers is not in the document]
dbazhal: You are right, 2 is too bad. Changed to None.

kopf/reactor/invocation.py
```diff
@@ -13,7 +13,9 @@
 
 # The executor for the sync-handlers (i.e. regular functions).
 # TODO: make the limits if sync-handlers configurable?
-executor = concurrent.futures.ThreadPoolExecutor(max_workers=3)
+from kopf.config import WorkersConfig
```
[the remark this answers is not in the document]
dbazhal: +

examples/04-events/example.py
```diff
@@ -1,6 +1,8 @@
 """
 Send the custom events for the handled or other objects.
 """
+import asyncio
```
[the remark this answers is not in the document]
dbazhal: removed

examples/99-all-at-once/example.py
```diff
@@ -1,7 +1,7 @@
 """
 Kubernetes operator example: all the features at once (for debugging & testing).
 """
-
+import asyncio
```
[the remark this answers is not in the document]
dbazhal: done

nolar:
As a side-note: I tried to check why Travis CI pipeline is not triggered for this PR, and for this PR only — it works for other PRs & contributors. Travis CI says "Abuse detected" without any additional explanation. 

I will try to contact Travis CI support to clarify what is the actual reason. 

My only assumption for now is that this is a PR from the "master" branch of a fork to the "master" branch of the upstream — maybe Travis does not like it. This is the only "unusual" thing in this PR. Just guessing.

kopf/config.py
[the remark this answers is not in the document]
nolar: So, these are Kopf's k8s-event levels, not Python's logging-levels. Now it is clear for me. Thanks.

kopf/k8s/events.py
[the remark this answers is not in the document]
dbazhal: As I understand, if you are already not in master thread, get_running_loop will raise RuntimeError. Same story with get_event_loop - it's for use in master thread, so, supposed you have only one event loop, and run all required tasks in it.
In current kopf architecture it is hard to create 1 event loop and run all kinds of tasks in it, so this is dirty hack to create new event loops inside functions that by themselves already run in slave threads.
It requires some controller object to be created before any of kopf functionality is initiated. So you could then create it, configure it, run/chain and control lifecycle of any required tasks(watching/queueing/handling/event-posting/object-patching) in it.
I understand that this current solution is dirty, but I don't know how to do it better at low costs right now.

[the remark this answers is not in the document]
dbazhal: Reading it again. Can't really understand how it works from docs. Will try ```get_running_loop``` later and comment here on results.

kopf/config.py
```diff
@@ -77,3 +83,35 @@ def configure(debug=None, verbose=None, quiet=None):
...
+    synchronous_handlers_threadpool_limit = None  # if None, calculated by ThreadPoolExecutor based on cpu count
+
+    """How long does a worker can idle before exiting and garbage-collecting."""
+    worker_idle_timeout = 5.0
+
+    """How fast/slow does a worker deplete the queue when an event is received."""
+    worker_batch_window = 0.1
+
+    """How long does a worker can work on watcher exit before being cancelled. """
+    worker_exit_timeout = 2.0
```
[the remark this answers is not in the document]
dbazhal: Ok, got it, fixed that, thank you.

tests/k8s/test_patching.py
```diff
@@ -10,7 +12,10 @@ def test_by_name_clustered(client_mock, resource):
     sidefn_mock = apicls_mock.return_value.patch_namespaced_custom_object
     mainfn_mock = apicls_mock.return_value.patch_cluster_custom_object
 
-    res = patch_obj(resource=resource, namespace=None, name='name1', patch=patch)
+    task = asyncio.create_task(patch_obj(resource=resource, namespace=None, name='name1', patch=patch))
+    loop = asyncio.get_event_loop()
+    loop.run_until_complete(task)
+    res = task.result
```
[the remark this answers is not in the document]
dbazhal: Done, thank you.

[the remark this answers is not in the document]
dbazhal: Ok, tests fixed.

kopf/k8s/events.py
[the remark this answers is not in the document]
dbazhal: Changed to ```get_running_loop```. Seems to work fine, tests are passing.

kopf/k8s/patching.py
dbazhal: Fixed this one too.

kopf/k8s/events.py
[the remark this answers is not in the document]
nolar: Thanks. 

To clarify: every thread has its own loop by default. `get_event_loop()` gets-or-creates it in the current thread. However, those are different event-loops for each thread, i.e. not the same loop as in the main thread.

`get_current_loop()` is only possible from inside a coroutine, as the coroutines already run in a loop (by definition). 

In case of Kopf, it will be always the main thread's event-loop (or a thread where the `run()` method is called), as other threads are used only for sync functions via the thread pools — unless we intentionally create and use a loop inside of those sync functions (but we don't).

Therefore, the effect of these lines is the same: it will be a main thread's event-loop. But a one-liner `get_current_loop()` is more simple and clear than the four-liner above.

kopf/reactor/queueing.py
nolar: Ah, got it! If there are no pending tasks, there is nothing to cancel, so there will be no cancelled tasks. But a list is anyway needed few lines below for exception re-raising.

kopf/k8s/events.py
```diff
@@ -1,15 +1,23 @@
...
+from kopf import config
+
 logger = logging.getLogger(__name__)
 
 MAX_MESSAGE_LENGTH = 1024
 CUT_MESSAGE_INFIX = '...'
 
 
-def post_event(*, obj, type, reason, message=''):
+event_executor = concurrent.futures.ThreadPoolExecutor(max_workers=config.WorkersConfig.synchronous_event_post_workers_limit)
```
nolar: Here is a new question. The intention of this PR is to make thread-pools' sizes configurable (and a lot of other things too).

However, if you `import kopf` (or `kopf run` in CLI), all the modules will be loaded, this one included, and the executors will be created with the default hard-coded config (as they are the module-level objects).

Any config changes applied after the import will have no effect on the executors.

What is the intended way of changing the configs to alter the Kopf's runtime setup?

nolar: So, after a quick discussion with the PR's author (in a chat), we decided to go as is for now. 

These things (thread pool sizes) were not configurable before the PR, so they will be not configurable after the PR — nothing changes here.

However, blocking operations (API calls) are now moved from asyncio event-loops to the thread pools — which gives a great performance gain already, and was the main goal in this PR.

And the introduced config objects can be used to dynamically configure the Kopf's runtime later, in a separate PR (since configurability is a separate topic).

docs/configuring.rst
```diff
@@ -0,0 +1,60 @@
...
+
+There are tools to configure some of kopf functionality, like asynchronous
+tasks behaviour and logging events.
+
+.. note::
+    All configuration should be done before importing kopf modules.
+
+
+Configure logging events
+=================
```
nolar: These "==="-lines should be at least the same length as the text they use as a caption/title: http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections

docs/configuring.rst
```diff
@@ -0,0 +1,60 @@
...
+
+* ``kopf.config.LOGLEVEL_INFO``
+* ``kopf.config.LOGLEVEL_WARNING``
+* ``kopf.config.LOGLEVEL_ERROR``
+* ``kopf.config.LOGLEVEL_CRITICAL``
+
+.. code-block:: python
+    :caption: test_example_operator.py
+
+    from kopf import config
```
nolar: Does it work? I tried with a local simulation:

```python
# pkg/__init__.py
print('init!')
import pkg.unwanted
import pkg.mod
```

```python
# pkg/mod.py
print('mod')
```

```python
# pkg/unwanted.py
print('UNWANTED')
```

```python-console
>>> from pkg import mod
init!
UNWANTED
mod
```

docs/configuring.rst
```diff
@@ -0,0 +1,60 @@
...
+
+    # Now kopf will send events only when error or critical occasion happens
+    config.EventsConfig.events_loglevel = config.LOGLEVEL_ERROR
```
nolar: Everything beside thread pool sizes, such as this loglevel, can be set at any time after the imports. The values are used as they are at the moment of their usage, they are never remembered locally. Or do I miss something?

docs/configuring.rst
dbazhal: Omg, I missed __init__ :) no, that does not work at all.

dbazhal: Trying to figure out how it could be configured.

nolar: @dbazhal I believe, it cannot. This is why I've made a comment above (the one with _"after a quick discussion"_) about skipping this topic for now, postponing it for another PR about configurability and isolated user-controlled event-loops.

One way would be to have `kopf_config` module outside of `kopf` package, as a top-level module. But that doesn't look nice. A better way would be to have a `cfg` instance object passed down the call stack starting from `run()` or `create_tasks()` — but it is a big changeset.

kopf/config.py
```diff
@@ -103,23 +105,37 @@ class WorkersConfig:
...
+            logging.debug('Setting up syn executor')
+            WorkersConfig.threadpool_executor = concurrent.futures.ThreadPoolExecutor(
+                max_workers=WorkersConfig.synchronous_tasks_threadpool_limit
+            )
+        return WorkersConfig.threadpool_executor
+
+    @staticmethod
+    def set_synchronous_tasks_threadpool_limit(new_limit: int):
+        if new_limit < 1:
+            return
```
nolar: Perhaps, `ValueError`.

docs/configuring.rst
dbazhal: fixed, thx

docs/configuring.rst
dbazhal: You are right, fixed that.

kopf/config.py
dbazhal: Your truth, ```ValueError``` it be :)

Issue #59, opened 2019-05-10
Handler for the operator start/stop


nolar:
To work properly with different Kubernetes client libraries, such as PyKube/PyKubeNG (#15), the authentication is needed.

One way to do this is on the module import level. Yet, this is wrong from the code structuring point of view.

It would be better to have the explicit event handler for the operator-wide (not just object-wide) events:

```python
import kopf
import kubernetes
import pykube

config = None

@kopf.on.startup()
def auth_pykube(**_):
    global config
    try:
        config = pykube.KubeConfig.from_service_account()  # cluster env vars
    except FileNotFoundError:
        config = pykube.KubeConfig.from_file()  # developer's config files

@kopf.on.startup()
def auth_client(logger, **_):
    try:
        kubernetes.config.load_incluster_config()  # cluster env vars
        logger.debug("configured in cluster with service account")
    except kubernetes.config.ConfigException as e1:
        try:
            kubernetes.config.load_kube_config()  # developer's config files
            logger.debug("configured via kubeconfig file")
        except kubernetes.config.ConfigException as e2:
            raise Exception(f"Cannot authenticate neither in-cluster, nor via kubeconfig.")

@kopf.on.cleanup()
def purge_caches(**_):
    pass
```

It could be also used to start some background threads or async coroutines for monitoring the unrelated systems (e.g. API polling).

The handler should be called before any implicit `login()` or other API calls, so that nothing is implicitly configured by the framework. But after the logging is configured and registry is populated — so that the handler itself is logged, but it could adjust the logging, or the registry.

For the exit-handlers, perhaps the [`atexit`](https://docs.python.org/3/library/atexit.html) stdlib can be used.

nolar:
Added in `kopf==0.23rc1` (see the [release notes](https://github.com/zalando-incubator/kopf/releases/tag/0.23rc1)). Also, custom authentication methods and `@kopf.on.login()` is added there.

Documentation:

* https://kopf.readthedocs.io/en/latest/handlers/#startup-handlers
* https://kopf.readthedocs.io/en/latest/handlers/#cleanup-handlers
* https://kopf.readthedocs.io/en/latest/authentication/

Examples:

* https://github.com/zalando-incubator/kopf/blob/master/examples/13-hooks/example.py

Issue #57, opened 2019-05-10
Implicitly maps kinds<=>plurals


nolar:
Currently, the resources' plural names must be used as the only acceptable reference to the resources in the handlers:

```
@kopf.on.create('zalando.org', 'v1', 'kopfexamples')
def create_fn(**_):
    pass
```

It would be more convenient to refer to the resource the same way as the YAML files do: by their "kind":

```
@kopf.on.create('zalando.org', 'v1', 'KopfExample')
def create_fn(**_):
    pass
```

Or, as even more convenient way, as in `kubectl` — by any of their names without group/version references:

```
@kopf.on.create('KopfExample')  # kind
def create_fn(**_):
    pass

@kopf.on.create('kopfexample')  # singular
def create_fn(**_):
    pass

@kopf.on.create('kopfexamples')  # plural
def create_fn(**_):
    pass
```

PR #110, opened 2019-06-14
Switch to pykube-ng to handle arbitrary resources (pods, jobs, etc)

### LINKED ISSUE #15: Consider pykube-ng?

### LINKED ISSUE #84: Support for built-in resources (pods, jobs, etc)
## Expected Behaviour

Kopf should be able to handle the built-in resources, like pods, jobs, pvcs, etc. — Both in the spy-handlers (#30) and in the regular handlers (if decided so).

## Actual Behaviour

Currently, Kopf is only able to serve the custom resources — because it uses the `kubernetes.client.CustomObjectsApi` API (for watching & for patching).

## Suggestions

~Maybe,~ pykube-ng can be beneficial here, as it has a nice syntax to operate on all the resources, built-in and custom. See #15. With the official Kubernetes client, we have to implement our own branching on the object kind, and using the proper class+methods for every kind — basically turning Kopf into a Kubernetes client library, which is not the goal.

~Maybe,~ a pre-scan of the cluster would be needed to identify all the existing resources. See #57.

## Specifications

```
kopf==0.13
```

### LINKED ISSUE #137: Can this framework be used to monitor core resources like namespaces?
## Expected Behavior
I want to handle namespaces events. Can this be done with kopf?

## Actual Behavior
```
[2019-07-08 15:42:29,693] kopf.clients.auth    [DEBUG   ] configured via kubeconfig file
[2019-07-08 15:42:30,561] kopf.engines.peering [WARNING ] Default peering object not found, falling back to the standalone mode.
Traceback (most recent call last):
  File "/usr/local/bin/kopf", line 10, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/kopf/cli.py", line 28, in wrapper
    return fn(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/kopf/cli.py", line 59, in run
    peering_name=peering_name,
  File "/usr/local/lib/python3.7/site-packages/kopf/reactor/queueing.py", line 271, in run
    task.result()  # can raise the regular (non-cancellation) exceptions.
  File "/usr/local/lib/python3.7/site-packages/kopf/reactor/queueing.py", line 78, in watcher
    async for event in watching.infinite_watch(resource=resource, namespace=namespace):
  File "/usr/local/lib/python3.7/site-packages/kopf/clients/watching.py", line 132, in infinite_watch
    async for event in streaming_watch(resource=resource, namespace=namespace):
  File "/usr/local/lib/python3.7/site-packages/kopf/clients/watching.py", line 81, in streaming_watch
    rsp = fetching.list_objs(resource=resource, namespace=namespace)
  File "/usr/local/lib/python3.7/site-packages/kopf/clients/fetching.py", line 64, in list_objs
    plural=resource.plural,
  File "/usr/local/lib/python3.7/site-packages/kubernetes/client/apis/custom_objects_api.py", line 1307, in list_cluster_custom_object
    (data) = self.list_cluster_custom_object_with_http_info(group, version, plural, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/kubernetes/client/apis/custom_objects_api.py", line 1413, in list_cluster_custom_object_with_http_info
    collection_formats=collection_formats)
  File "/usr/local/lib/python3.7/site-packages/kubernetes/client/api_client.py", line 334, in call_api
    _return_http_data_only, collection_formats, _preload_content, _request_timeout)
  File "/usr/local/lib/python3.7/site-packages/kubernetes/client/api_client.py", line 168, in __call_api
    _request_timeout=_request_timeout)
  File "/usr/local/lib/python3.7/site-packages/kubernetes/client/api_client.py", line 355, in request
    headers=headers)
  File "/usr/local/lib/python3.7/site-packages/kubernetes/client/rest.py", line 231, in GET
    query_params=query_params)
  File "/usr/local/lib/python3.7/site-packages/kubernetes/client/rest.py", line 222, in request
    raise ApiException(http_resp=r)
kubernetes.client.rest.ApiException: (404)
Reason: Not Found
HTTP response headers: HTTPHeaderDict({'Audit-Id': '818d48ae-127f-4dc7-b767-337cc49be806', 'Content-Type': 'text/plain; charset=utf-8', 'X-Content-Type-Options': 'nosniff', 'Date': 'Mon, 08 Jul 2019 22:42:30 GMT', 'Content-Length': '19'})
HTTP response body: 404 page not found
```

  1. Create sample code sample.py:
```
import kopf
@kopf.on.create('core', 'v1', 'namespaces')
def module_handler(meta, spec, namespace, logger, **kwargs):
    logger.info(f"Namespace was created")
```
  2. Run code `kopf run sample.py --verbose`
  

## Specifications

  - Platform:
  - Kubernetes version: v1.12.6
  - Python version: 3.7.1
  - Python packages installed: 
aiohttp==3.5.4
aiojobs==0.2.2
async-timeout==3.0.1
attrs==19.1.0
awscli==1.16.169
boto3==1.9.134
botocore==1.12.159
cachetools==3.1.1
certifi==2019.6.16
chardet==3.0.4
Click==7.0
colorama==0.3.9
docutils==0.14
google-auth==1.6.3
idna==2.8
iso8601==0.1.12
jmespath==0.9.4
kopf==0.16
kubernetes==9.0.0
multidict==4.5.2
oauthlib==3.0.1
packaging==19.0
pip==19.1.1
pip-review==1.0
pyasn1==0.4.5
pyasn1-modules==0.2.5
pyparsing==2.4.0
python-dateutil==2.8.0
PyYAML==3.13
requests==2.22.0
requests-oauthlib==1.2.0
rsa==3.4.2
s3transfer==0.2.0
setuptools==41.0.0
six==1.12.0
urllib3==1.24.3
websocket-client==0.56.0
wpm==1.51.4
yarl==1.3.0

```

### PR DESCRIPTION
**Pull request by nolar**
> Issue : closes #15, closes #84, closes #137, related #97 (docs).

## Description

Switch the internal K8s client library from the official library `kubernetes` to [`pykube-ng`](https://github.com/hjacobs/pykube).

This simplifies handling of arbitrary resource types, such as pods, jobs, etc — for any handlers: both regular handlers (`@kopf.on.create/update/delete/resume`), and spy-handles (`@kopf.on.event`).

The previous Kubernetes client library is supported _optionally_: if present, it will be auto-logged in, to keep compatibility with the previous behaviour. Otherwise, it is ignored. And it is not used anywhere. It will be removed from 1.0 release.

## Types of Changes

- Refactor/improvements

**PLEASE NOTE:** The K8s client library used inside of Kopf is Kopf's internal implementation detail. Only the native data structures (dicts, lists, etc) are exposed from the framework to the operator and accepted back. The operators MUST NOT rely on how Kopf talks to the K8s APIs. It can be `requests`/`requests-async` in the future (just as an example). The only dependency is auto-login — to be removed with #59 (the operators must explicitly authenticate themselves; now they can do this on the module-level code only).

## Todos

* [ ] Add and examples, revert examples 01/02 to what they were initially.
* [ ] Extend the docs to point that arbitrary resources are supported.
* [ ] Ensure it works in all cases with all features and nothing is broken (a real test-drive for 1-2 weeks).

## Review
- [ ] Tests
- [ ] Documentation

### CODE REVIEW

**Comment by angelbarrera92** [comment]
I tried this PR locally on K3s and i was having connectivity problems.
For those who want to try it on k3s locally (fedora), please go to the following k3s issue: https://github.com/rancher/k3s/issues/24

Thanks @nolar for your support!

File: kopf/k8s/classes.py [hunk 1]
```diff
@@ -0,0 +1,48 @@
+import json
+from typing import Type
+
+import pykube
+
+from kopf.k8s import config
+
+
+# TODO: this mixin has to be migrated into pykube-ng itself.
```
hjacobs: feel free to open a PR for pykube-ng :smile:

kopf/k8s/classes.py
```diff
@@ -0,0 +1,48 @@
...
+    pass
+
+
+class CustomResourceDefinition(APIObject):
+    version = "apiextensions.k8s.io/v1beta1"
+    endpoint = "customresourcedefinitions"
+    kind = "CustomResourceDefinition"
+
+
+def _make_cls(resource) -> Type[APIObject]:
```
hjacobs: I guess this can be part of pykube-ng, too

kopf/k8s/config.py
```diff
@@ -0,0 +1,134 @@
...
+        import kubernetes
+    except ImportError:
+        pass
+    else:
+        login_client()
+
+
+def login_pykube():
+    global _pykube_cfg
+    try:
```
hjacobs: I guess we can add some "auto.." method for pykube-ng itself to not repeat the logic everywhere (I do the same as you in my projects: first try "in-cluster" then try kubeconfig file)

nolar:
All pre-requisites are now in master. This branch is rebased on top of master, with few fixes here & there. The implementation is mainly the same.

**Ready for review** — to be released as a pre-release version (presumably `0.19rc1`), and to be tested in our own apps before going to the public release.

kopf/k8s/classes.py
nolar: I will for sure. First of all, let's see how it works in this mode. And second:

With the official client, there was an issue when the switch the Content-Type from merge-patch format (dicts of fields) to JSON-patch format (lists of ops) — see #134 and kubernetes-client/python#866.

For Kopf, merge-patch is needed and is sufficient. But for a generic k8s client, both formats should be supported.

hjacobs:
@nolar I created a PR to move reasonable changes upstream to `pykube-ng`: https://github.com/hjacobs/pykube/pull/29

examples/requirements.txt
```diff
@@ -1,3 +1,4 @@
 kopf
 kubernetes
```
thilp: As far as I can see, the only remaining use of `kubernetes` in `examples/` is in `02-children/example.py`, where it could easily be replaced with Pykube as well. Is it on purpose, to show that Kopf can be used with both libraries?

kopf/clients/auth.py
```diff
@@ -1,42 +1,125 @@
...
+
+    Automatic refresh/reload of the tokens or objects also should be done here.
     """
 
-    # Configure the default client credentials for all possible environments.
+    # Pykube login is mandatory. If it fails, the framework will not run at all.
+    try:
+        import pykube
+    except ImportError:
+        raise  # mandatory
```
thilp: `pykube` is now imported at the module-level, I suppose this can be removed.

kopf/clients/auth.py
```diff
@@ -1,42 +1,125 @@
...
+    else:
+        login_pykube(verify=verify)
+
+    # We keep the official client library auto-login only because it was
+    # an implied behavior before switching to pykube -- to keep it so (implied).
```
thilp: Is there some kind of deprecation/removal planned for this?

kopf/clients/auth.py
```diff
@@ -1,42 +1,125 @@
...
+    try:
+        api = get_pykube_api()
+        rsp = api.get(version="", base="/")
+        rsp.raise_for_status()
+    except requests.exceptions.HTTPError as e:
+        if e.response.status_code == 401:
+            raise AccessError("Cannot authenticate to the Kubernetes API. "
+                              "Please login or configure the tokens.")
+        else:
+            raise
```
thilp: I was surprised to get a `403` from there the other day. I eventually discovered it was because my operator's service account was missing the following role:
```yaml
nonResourceURLs: ["/"]
verbs: ["get"]
```
(This was never a problem when Kopf was using the `kubernetes` client.)
Do you think it would be worth catching and mentioning here (not necessarily in this PR)?

kopf/clients/auth.py
```diff
@@ -1,42 +1,125 @@
...
+
+def get_pykube_cfg() -> pykube.KubeConfig:
+    if _pykube_cfg is None:
+        raise LoginError("Not logged in with PyKube.")
+    return _pykube_cfg
+
+
+# TODO: add some caching, but keep kwargs in mind. Maybe add a key= for purpose/use-place?
+def get_pykube_api(timeout=None) -> pykube.HTTPClient:
+    return pykube.HTTPClient(get_pykube_cfg(), timeout=timeout)
```
thilp: What about specifying a non-infinite timeout by default? Pykube's default is 10s, but I think explicitly passing `None` here would deactivate that reasonable default.

kopf/clients/classes.py
```diff
@@ -0,0 +1,48 @@
...
+    kind = "CustomResourceDefinition"
+
+
+def _make_cls(resource) -> Type[APIObject]:
+    api = auth.get_pykube_api()
+    api_resources = api.resource_list(resource.api_version)['resources']
+    resource_kind = next((r['kind'] for r in api_resources if r['name'] == resource.plural), None)
+    is_namespaced = next((r['namespaced'] for r in api_resources if r['name'] == resource.plural), None)
+    if not resource_kind:
+        raise pykube.ObjectDoesNotExist("No such CRD at all.")
```
thilp: This exception could be made more user-friendly by specifying the resource kind that was requested. Perhaps something like this?
```suggestion
        raise pykube.ObjectDoesNotExist(f"No such CRD: {resource.plural!r}")
```

kopf/logging/logging.py
```diff
@@ -0,0 +1,41 @@
...
+    try:
+        import kubernetes
+    except ImportError:
+        pass
+    else:
+        config = kubernetes.client.configuration.Configuration()
+        config.logger_format = format
+        config.logger_file = None  # once again after the constructor to re-apply the formatter
+        config.debug = debug
+        kubernetes.client.configuration.Configuration.set_default(config)
```
thilp: This block already appears in `kopf/config.py`. I'm not sure how both relate. Could it be a duplicate?

tools/kubernetes-client.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -eu
+set -x
+
+if [[ "${CLIENT:-}" = "yes" ]] ; then
+    # FIXME: See https://github.com/kubernetes-client/python/issues/866
+    pip install 'kubernetes<10.0.0'
+fi
```
thilp: Where/how is this script used? I could find no references to it in the repository.

kopf/clients/classes.py
```diff
@@ -0,0 +1,48 @@
+import json
+from typing import Type
+
+import pykube
+
+from kopf.clients import auth
+
+
+# TODO: this mixin has to be migrated into pykube-ng itself.
+class APIObject(pykube.objects.APIObject):
```
hjacobs: I created https://github.com/hjacobs/pykube/pull/29 to have this upstream

kopf/clients/classes.py
```diff
@@ -0,0 +1,48 @@
...
+        ))
+        self.api.raise_for_status(r)
+        self.set_obj(r.json())
+
+
+class NamespacedAPIObject(pykube.objects.NamespacedAPIObject, APIObject):
+    pass
+
+
+class CustomResourceDefinition(APIObject):
```
hjacobs: I created https://github.com/hjacobs/pykube/pull/29 to have this upstream

kopf/clients/auth.py
hjacobs: Good catch, I think requesting `/` might be tricky to not forget in RBAC setups

kopf/clients/auth.py
nolar: I thought about the future when pykube-ng could also be replaced with something, e.g. with the raw `requests-async` — for the sake of async HTTP calls. Total asynchronicity is one of the mid-term goals (not a priority at the moment though, as it is "solved" with the thread-pools). In that case, these libraries can be used to parse the config/envvars only, to retrieve the token(s). Or not used at all.

kopf/clients/auth.py
nolar: Not decided on this yet. Definitely before 1.0 release. 

The deprecation of auto-login can be done only after there is a way (read: a handler) to login the operator itself when it starts — e.g. as in #59. Now, it can be done with the module-level code, but it looks hacky.

kopf/clients/auth.py
nolar: Good point. Maybe, there should be no verification at all. It should fail on the actual operations attempted. 

Looking wider, it can be so that authentication succeeds, but the authorization fails on specific resources or specific REST operations of those resources, such as `PATCH` (because of RBAC). The operator should perform a full self-check on startup. It will only cover the permissions needed by the framework (watching-patching), but will not cover the permissions needed by the operator's logic (e.g. creating pods/jobs). 

See also: #49 for such self-checks (against a RBAC yaml, not against the real cluster, but it can be extended).

Meanwhile, I will make it here to ignore all errors except for HTTP 401 and TCP/HTTPS connection errors — since the purpose is to verify authentication, not authorization.

tools/kubernetes-client.sh
nolar: It is part of CI builds, used in `.travis.yaml`.

examples/requirements.txt
nolar: Yes. And also to make sure the official client works (all examples are used as the e2e tests).

examples/10-builtins/README.md
```diff
@@ -0,0 +1,43 @@
...
+Since `kubectl run` creates a Deployment, not just a Pod,
+a new pod will be created every 30 seconds. Observe with:
+
+```bash
+kubectl get pods --watch
+```
+
+*Please note that Kopf puts a finalizer on the managed resources,
+so the pod deletion will be blocked unless the operator is running
+(to remove the finalizer). This will be made optional in #24.*
```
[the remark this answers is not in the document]
nolar: Removed.

kopf/clients/auth.py
nolar: Done in 2118b8ed0f9f06822ba694a4751784804267d8dc. Now, only connection errors and 401. All other errors are ignored.

kopf/clients/auth.py
nolar: Fixed.

kopf/clients/classes.py
nolar: Fixed. Just, the full name from `resource.name` is used — as it is seen in `kubectl get crd` — e.g. `kopfexamples.zalando.org`.

kopf/logging/logging.py
nolar: `kopf.logging` is a remnant of some early implementations, which has accidentally sneaked into the codebase when rebasing. It should not be here at all (and no other module use it) — so, removed.

kopf/clients/classes.py
hjacobs: Released in https://github.com/hjacobs/pykube/releases/tag/0.24

kopf/clients/classes.py
nolar: @hjacobs I've switched to `pykube-ng>=0.25`. So, the patching & CRD hacks are gone (including these APIObject/NamespacesAPIObject classes). See f105def0738a0e7dc43804ea7e4cf9cc9fff5a60

kopf/k8s/classes.py
nolar: @hjacobs In theory, there is already an implementation in `pykube.object_factory()`, and it is an officially documented way of creating the classes.

But I had issues with it. First, the self-made mixins with the `patch()` method (now solved). Second, because it requires `kind` as a search criterion, and Kopf uses the canonical plural names from the `name` field (as in the K8s API URLs, and as in the original Kubernetes client library).

The automatic implicit mapping kubectl-style (i.e. any form is accepted, including the short-names) is planned in #57, and it will be more convenient for the users, but it is not yet there.

I would be nice if pukube-ng could search by any name field (kind, plural, singular, all shortnames) — same as kubectl does. But it can be done as a separate PR in Kopf.

nolar:
@thilp @hjacobs All issues are addressed (I think). Can be re-reviewed. Please, comment/highlight again if I missed something.

kopf/clients/auth.py
```diff
@@ -82,22 +82,36 @@ def login_client(verify=False):
...
+
+    All other errors (e.g. 403, 404) are ignored: it means, the host and port
+    are configured and are reachable, the authentication token is accepted,
+    and the rest are authorization or configuration errors (not a showstopper).
     """
     try:
         api = get_pykube_api()
         rsp = api.get(version="", base="/")
         rsp.raise_for_status()
+        api.raise_for_status(rsp)  # replaces requests's HTTPError with its own.
```
thilp: I think the `rsp.raise_for_status()` can be removed, because `api.raise_for_status(rsp)` [already does it](https://github.com/hjacobs/pykube/blob/dcf39979bc336dd36bb5f70f993797c7d19741a5/pykube/http.py#L239-L241) and you are already catching everything.

Basically if the first line raises, the second would as well (but will not be run because the first raised), and if the first line doesn't raise, then the second won't either, so nothing is gained by having both.

nolar:
Released as [`kopf==0.19rc1`](https://github.com/zalando-incubator/kopf/releases/tag/0.19rc1). 

Note: it is a pre-release, so it requires either `pip install --pre kopf`, or explicit version pinning: `pip install kopf==0.19.rc1`.

kopf/clients/auth.py
nolar: This behaviour caused some issues with 0.19rc1 and will be reverted to no-timeouts for all internal api calls, and especially the watch-requests — same as it was before migration to pykube: https://github.com/zalando-incubator/kopf/pull/144

Issue #51, opened 2019-04-30
Freeze for 2+ operators of the same priority


nolar:
## Current Behaviour

When 2+ operators with the same priority run (e.g. default of 0), both are working. Which breaks the purpose of the peering to prevent duplicated side-effects.

## Expected Behaviour

When 2+ operators with the same priority run, with peering configured and functional, only one should be functioning, others should freeze.

The best case would be that all operators issue a warning and freeze, so that the cluster becomes not served anymore — this will bring people attention sooner than if they negotiate which one should work. 2+ operators with the same priority is obviously a deployment error.

_Note: If the operators make a decision on which one runs, which one freezes (e.g. by the most recent pod start time, or by id, or anything else comparable), then it could happen that the old deployment in another namespace serves the cluster, when a new operator is expected to work. It is better to stop serving and bring people attention to resolve the situation._

## Suggestion

See [the module where it all happens](https://github.com/zalando-incubator/kopf/blob/d1cb7331b42255f1610e57e3855db203736bcacc/kopf/reactor/peering.py#L206-L215).

## Checklist

* [ ] Implementation.
* [ ] Tests — manually (there is no auto-tests for peering yet).
* [ ] Documentation (in `docs/peering.rst`)

smileisak:
@nolar is this issue is already treated ?

nolar:
@smileisak To my knowledge, no, it is free and waiting.

PR #114, opened 2019-06-14
Freeze all operators with same priority and warn

smileisak:
# One-line summary

> Issue : #51

## Description

When two operators runs with the same priority, both will handle the same request. This PR adds the possibility to freeze all events with the same priority, as mentioned in #51 and print a warning error.

## Types of Changes

## Review
- [x] Manual Tests
- [x] Documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,9 @@ junit.xml
 # Documentation
 docs/_build
 docs/packages
+
+# VirtualEnv
+env
+
+# VSCode
+.vscode
--- a/docs/peering.rst
+++ b/docs/peering.rst
@@ -14,7 +14,8 @@ notices that other operators start with a higher priority, it freezes
 its operation until those operators stop working.
 
 This is done to prevent collisions of multiple operators handling
-the same objects.
+the same objects. If two operators runs with the same priority  all operators 
+issue a warning and freeze, so that the cluster becomes not served anymore.
 
 To set the operator's priority, use :option:`--priority`:
 
--- a/kopf/engines/peering.py
+++ b/kopf/engines/peering.py
@@ -40,12 +40,8 @@ from typing import Iterable, Mapping, Optional, Union
 
 import iso8601
 
-<<<<<<< HEAD:kopf/engines/peering.py
 from kopf.clients import fetching
 from kopf.clients import patching
-=======
-from kopf.k8s import fetching, patching
->>>>>>> All operators with the same priority issue a warning and freeze, so that the cluster becomes not served anymore:kopf/reactor/peering.py
 from kopf.reactor import registries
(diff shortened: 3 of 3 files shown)


kopf/reactor/peering.py
```diff
@@ -231,12 +229,16 @@ def apply_peers(
...
             logger.warning(f"Possibly conflicting operators with the same priority: {same_peers}.")
-        if freeze.is_set():
-            logger.info(f"Resuming operations after the freeze.")
-            freeze.clear()
+            logger.warning(f"Freezed all Operators: {peers}")
+            freeze.set()
+        else:
+            if freeze.is_set():
+                logger.info(f"Resuming operations after the freeze. Conflicting operators with the same priority are gone")
+                freeze.clear()
```
nolar: Can this be done via one-level `if` (just to make it more readable; the logic is fine)?

```python
if prio_peers:
    ...
elif same_peers:
    ...
elif freeze.is_set():
    ...
```

Previously, there was a 2-step block (if same_peers + if freeze.is_set()). Now, it is either-or, so the nested if's are not needed.

smileisak: Sure ! I agree. Code Updated !

smileisak:
@samurang87 let me know if i need to change something in this PR. 

Thank you

nolar:
@smileisak Can you please resolve conflicts with the new master branch? (Or maybe give me the write access to your fork.) — And then I will merge it.

I guess, it is only the move `kopf.reactor.peearing` -> `kopf.engines.peering` here (see #124).

smileisak:
@nolar this breaks the CICD Pipeline, i sent you an invitation to have write access to my fork.

nolar:
@smileisak Oh! That's an interesting issue. It seems, it existed there since the beginning, and was only revealed now.

When an operator exits, it does not remove its record from the peering object. When the next operator starts fast enough (which is the case for the e2e tests), it believes that another operator is running, and freezes.

Previously, it was not a problem, since all test-operators were same-ranked, and the freeze was ignored (though the potential conflict was logged).

This is caused by how asyncio task cancellation works: it is a `CancelledError` exception injected into  a task (expected), but it also implicitly cancels all other coroutines/tasks created after the injection (unexpected) — and so, the peering self-deletion was also silently cancelled.

Now, with this new commit, the self-deletion is ["shielded"](https://docs.python.org/3/library/asyncio-task.html#asyncio.shield) from the cancellation of its calling task. And also logs if another exception has happened.

nolar:
Released in [`kopf==0.18`](https://github.com/zalando-incubator/kopf/releases/tag/0.18)

@smileisak Thanks once again for your contribution.

PR #128, opened 2019-07-03
Log to k8s events

nolar:
> Issue : #136 

_As a logical continuation of #125:_

## Description

Kopf is not a K8s API client library. Therefore, it should not provide any means of talking to the K8s API. The operators developers should use the K8s API client libraries of their choice. This also applies to K8s-event posting.

However, Kopf provides few routines for k8s-event posting: `kopf.event()`, `kopf.info()`, `kopf.warn()`, `kopf.exception()`. This is highly useful for logging not only to the logs (stdout/stderr), but also to K8s itself.

To deal with this dilemma, this PR changes the way how Kopf deals with this specific field — logging to k8s events:

Instead of providing the methods to explicitly post everything in addition to regular logging (thus duplicating the messages and code), everything logged to the per-object logger is now posted to K8s events transparently.

Was:

```python
def create_fn(logger, **kwargs):
    child = _create_child()
    kopf.info("Created a child: %s".format(child.id))
    logger.info("Created a child: %s", child.id)
```

Now:

```python
def create_fn(logger, **kwargs):
    child = _create_child()
    logger.info("Created a child: %s", child.id)
```

In the latter case, an k8s-event will be created too.

Only the log messages with level INFO+ will be posted. The DEBUG messages will **not** be posted (but will be logged as usually).

Only the messages logged to the per-object logger are posted. The rest of the `logging` machinery is **not** affected.

The existing k8s-event-posting routines remain in place. They continue to post the events as usually, attached to the specified object (their first positional argument).

All of the internal Kopf-originated events are now also posted that way — via the per-object logger (some were lowered to the DEBUG level to reduce noise).

## Types of Changes

- Refactor/improvements

## Review
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/kopf/__init__.py
+++ b/kopf/__init__.py
@@ -22,9 +22,6 @@ from kopf.config import (
     EventsConfig,
     WorkersConfig,
 )
-from kopf.engines.logging import (
-    ObjectLogger,
-)
 from kopf.engines.posting import (
     event,
     info,
--- a/kopf/__init__.py
+++ b/kopf/__init__.py
@@ -20,7 +20,10 @@ from kopf.config import (
     LOGLEVEL_ERROR,
     LOGLEVEL_CRITICAL,
     EventsConfig,
-    WorkersConfig
+    WorkersConfig,
+)
+from kopf.engines.logging import (
+    ObjectLogger,
 )
 from kopf.engines.posting import (
     event,
--- a/kopf/engines/logging.py
+++ b/kopf/engines/logging.py
@@ -48,20 +48,24 @@ class K8sPoster(logging.Handler):
         return has_ref and not skipped and super().filter(record)
 
     def emit(self, record):
-        type = (
-            "Debug" if record.levelno <= logging.DEBUG else
-            "Normal" if record.levelno <= logging.INFO else
-            "Warning" if record.levelno <= logging.WARNING else
-            "Error" if record.levelno <= logging.ERROR else
-            "Fatal" if record.levelno <= logging.FATAL else
-            logging.getLevelName(record.levelno).capitalize())
-        reason = 'Logging'
-        message = self.format(record)
-        self.queue.put_nowait(posting.K8sEvent(
-            ref=record.k8s_ref,
-            type=type,
-            reason=reason,
-            message=message))
+        # Same try-except as in e.g. `logging.StreamHandler`.
+        try:
+            type = (
+                "Debug" if record.levelno <= logging.DEBUG else
+                "Normal" if record.levelno <= logging.INFO else
+                "Warning" if record.levelno <= logging.WARNING else
+                "Error" if record.levelno <= logging.ERROR else
+                "Fatal" if record.levelno <= logging.FATAL else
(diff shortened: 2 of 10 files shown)


kopf/engines/logging.py
```diff
@@ -0,0 +1,141 @@
...
+    A logger/adapter to carry the object identifiers for formatting.
+
+    The identifiers are then used both for formatting the per-object messages
+    in `ObjectPrefixingFormatter`, and when posting the per-object k8s-events.
+
+    Constructed in event handling of each individual object.
+
+    The internal structure is made the same as an object reference in K8s API,
+    but can change over time to anything needed for our internal purposes.
+    However, the as little information should be carried as possible,
```
thilp:
```suggestion
    However, as little information should be carried as possible,
```

tests/posting/test_log2k8s.py
```diff
@@ -0,0 +1,88 @@
...
+
+OBJ1 = {'apiVersion': 'group1/version1', 'kind': 'Kind1',
+        'metadata': {'uid': 'uid1', 'name': 'name1', 'namespace': 'ns1'}}
+REF1 = {'apiVersion': 'group1/version1', 'kind': 'Kind1',
+        'uid': 'uid1', 'name': 'name1', 'namespace': 'ns1'}
+
+
+@pytest.fixture()
+def logstream(caplog):
+    """ Prefixing is done at the final output. We have to intercept it. """
```
thilp: This fixture is almost identical to the one in `test_cause_logging.py`, except it sets `logger.handlers` back to its old value after the test (whereas the other fixture sets `logger.handlers` to the empty list afterwards). Why this difference?

kopf/__init__.py
```diff
@@ -20,7 +20,10 @@
     LOGLEVEL_ERROR,
     LOGLEVEL_CRITICAL,
     EventsConfig,
-    WorkersConfig
+    WorkersConfig,
+)
+from kopf.engines.logging import (
+    ObjectLogger,
```
thilp: Why export this type at such a high level? It is not used from the root `kopf` module internally and, if I understand correctly, the only interactions the user could/should have with it are:
- use the instance already provided as the `logger` parameter in handlers (so no need to call the constructor),
- in type annotations, where encouraging users to rather use the _Logger_  or _LoggerAdapter_ interfaces would give Kopf more freedom for future refactorings.

nolar: You are right, let's not expose it to the public interface until there is a use-case. It is highly doubtful if anyone will instantiate their own per-object loggers at all. And it depend on the event-posting asyncio queue instance, which is surely not available to the users (it is part of Kopf's internals).

I will remove it from here.

tests/posting/test_log2k8s.py
nolar: Overlooked that. Fixed. Now, the fixture is in top-level `conftest.py`.

kopf/__init__.py
nolar: Removed.

Issue #119, opened 2019-06-25
Status handling


nolar:
Currently, status handling uses the `.status` stanza of the resource itself. Soon enough, a new method of working with `/status` sub-resource will be enabled by default: https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#subresources

Key differences:

* Status moves from `.status` field to `/status` sub-resource.
* PUT requests to the `/status` sub-resource ignore changes to anything except the status stanza.
* PUT/POST/PATCH requests to the custom resource ignore changes to the status stanza.

Things have to be refactored to make it work with the new mode.

Specifically:

* `.status` should not be taken into account when remembering the last-handled state. Different controllers can put information on `/status` sub-resource.
* All structures needed for handling the state of the operators should be moved out from `.status` to some other place. E.g. as used for "reporting" the status of the children objects (pods, jobs) and as "noticed" by the parent object (any kind of orchestration).
* Kopf actively uses status to store its own state in `.status.kopf`, and it also uses the status changes to trigger the next step of the event handling cycle. This should be re-worked to trigger the next cycle without the extra watch-events from the K8s API, as there can be none _(TODO: verify)_. Still, the state should be fully persistent and recoverable after every handler — for the case of process termination at any point (succeeded handlers should not re-run).

Issue #64, opened 2019-05-17
`metadata.generation` is reported in the diffs on post-creation


nolar:
## Expected Behavior

`metadata.generation` field is ignored in the last-seen state and diff calculation.

## Actual Behavior

`metadata.generation` is reported as a normally changed field and triggers the update handlers immediately on creation — in Minikube demos, where the generation is increased on every patch.

```
[2019-05-17 12:32:07,152] kopf.reactor.handlin [INFO    ] [default/kopf-example-1] All handlers succeeded for creation.
[2019-05-17 12:32:07,173] kopf.reactor.handlin [DEBUG   ] [default/kopf-example-1] Patching with: {'status': {'kopf': {'progress': None}, 'create': {'field': 'item1, item2'}}, 'metadata': {'annotations': {'kopf.zalando.org/last-handled-configuration': '{"apiVersion": "zalando.org/v1", "kind": "KopfExample", "metadata": {"annotations": {}, "generation": 1, "labels": {"somelabel": "somevalue"}, "name": "kopf-example-1", "namespace": "default"}, "spec": {"duration": "1m", "field": "value", "items": ["item1", "item2"]}}'}}}
[2019-05-17 12:32:07,296] kopf.reactor.handlin [DEBUG   ] [default/kopf-example-1] Update event: (('change', ('metadata', 'generation'), 1, 2),)
[2019-05-17 12:32:07,297] kopf.reactor.handlin [INFO    ] [default/kopf-example-1] All handlers succeeded for update.
```

## Specifications

  - Kubernetes version: Minikube + K8s 1.14.1

```
Client Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.0", GitCommit:"0ed33881dc4355495f623c6f22e7dd0b7632b7c0", GitTreeState:"clean", BuildDate:"2018-09-28T15:20:58Z", GoVersion:"go1.11", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.1", GitCommit:"b7394102d6ef778017f2ca4046abbaa23b88c290", GitTreeState:"clean", BuildDate:"2019-04-08T17:02:58Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"linux/amd64"}
```

  - Python packages installed: *(use `pip freeze --all`)*

```
kopf==0.10
```

PR #131, opened 2019-07-03
Purge status & metadata from the last-handled state

### LINKED ISSUE #129: last-handled-configuration should not include object status
## Expected Behavior
Kopf should not take object's status into account when making decision, should it process that object again or not, so there should no last-handled-configuration should not include object's status and not take it into account when making decision if object is changed. Kopf should be triggered only to object's metadata/spec changes.

## Actual Behavior
If a kopf operator is watching some kind of objects, and object status get's updated not only by kopf logic(status.X where X - handler id + status.kopf), kopf controller is then triggered to process that object, despite there is no actual changes in object itself. So there will be infinite process loop, where operator just update object's status back and forth.

  1. Write simple handler.
  1. Monkeypatch ```status.store_success = my_store_success```
  1. Make ```my_store_success``` do something like ```patch.setdefault('status', {})['conditions'] = result['conditions']```
  1. Run kopf operator and watch how it infinitely processes same objects.

## Specifications

Any.

nolar:
> Issue : closes #129, prepares #119, regression of #64, closes #130 (supersedes it).

## Description

Kopf remembers the last-handled state in the annotations. It is then used to detect if the object was handled previous and should be handled now, and also whether it is a CREATE or UPDATE cause (affects the selected handlers).

**`.status`:** Currently, the `.status` stenza is fully included in the last-handle state, except for Kopf's own structures (`.status.kopf`). According to the K8s conventions, status should be generally ignored when triggering the events, as all other controllers and operators can store any arbitrary information there. See also #119.

**`.metadata`:** Similarly for `.metadata` — new fields appear over time. For example, now it is `.metadata.deletionGracePeriodSeconds` that was added and sneaked into diffs, and caused the `@kopf.on.update` handlers. Kopf cannot follow all new fields in all Kubernetes versions. Instead, it will (since this PR) ignore all of metadata except for labels and annotations (except for surely useless annotations, like Kopf's own ones and kubectl's one). See also #64. 

**`.apiVersion/.kind`:** All resource kind references, including its name and namespace, are also excluded from the last-handled state. They are immutable, so it makes no sense to trace them.

**`.spec`:** All other stenzas like `.spec`, so as all other top-level fields (as in some resources) are treated as useful data even if not listed explicitly. They get to the last-handled state normally, always.

This PR does the following:

* Removes all of the system information (i.e. status & metadata & apiVersion & kind) from the remembered last-handled state.
* Re-adds labels & annotations.
* Re-adds the fields explicitly tracked in the field-handlers. Example:

```python
@kopf.on.field('zalando.org', 'v1', 'kopfexamples', field='status.expected-field')
def expected_field_changed(old, new, diff, **kwargs):
     print(diff)
```

In this example, the status sub-field will be taken into the last-handled state dumps because it is explicitly mentioned. All other status fields will be ignored.

## Side-effects

The old & new states, so as the diff structure will be provided to the creation and deletion handlers. Not much use of them there, so it will probably affect nothing. Previously, the old/new/diff kwargs were passed with `None`.

## Types of Changes

## Review
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/kopf/reactor/causation.py
+++ b/kopf/reactor/causation.py
@@ -83,6 +83,7 @@ def detect_cause(
     which performs the actual handler invocation, logging, patching,
     and other side-effects.
     """
+    diff = kwargs.get('diff')
     body = event['object']
     initial = event['type'] is None  # special value simulated by us in kopf.reactor.watching.
 
@@ -129,7 +130,7 @@ def detect_cause(
 
     # Cases with no state changes are usually ignored (NOOP). But for the "None" events,
     # as simulated for the initial listing, we call the resuming handlers (e.g. threads/tasks).
-    if not lastseen.is_state_changed(body) and initial:
+    if not diff and initial:
         return Cause(
             event=RESUME,
             body=body,
@@ -138,7 +139,7 @@ def detect_cause(
 
     # The previous step triggers one more patch operation without actual changes. Ignore it.
     # Either the last-seen state or the status field has changed.
-    if not lastseen.is_state_changed(body):
+    if not diff:
         return Cause(
             event=NOOP,
             body=body,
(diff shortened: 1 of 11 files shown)


dneuhaeuser-zalando:
There are conflicting files. At least zappr requires such conflicts to be resolved before review.

nolar:
@dneuhaeuser-zalando Resolved.

PR #133, opened 2019-07-03
Add file's dir to sys.path

### LINKED ISSUE #92: kopf disallow importing modules on path

nolar:
# One-line summary
Add current dir to sys.path prior to imports

> Issue : closes #92 
> Origin: closes #93 (suggestions added and conflicts resolved; original author: @trondhindenes)

## Description
Fixes the bug where modules weren't importable from current path since `current dir` was missing from `sys.path` (where modules are resolved from)

## Types of Changes

### DIFF
--- a/kopf/utilities/loaders.py
+++ b/kopf/utilities/loaders.py
@@ -14,9 +14,11 @@ Currently, two loading modes are supported, both are equivalent to Python CLI:
 Multiple files/modules can be specified. They will be loaded in the order.
 """
 
+import sys
 import importlib
 import importlib.util
 import os.path
+
 
 
 def preload(paths, modules):
@@ -25,6 +27,8 @@ def preload(paths, modules):
     """
 
     for path in paths:
+        print(path  )
+        sys.path.append(os.path.abspath(os.path.join(path, os.pardir)))
         name, _ = os.path.splitext(os.path.basename(path))
         spec = importlib.util.spec_from_file_location(name, path)
         module = importlib.util.module_from_spec(spec)
--- a/kopf/utilities/loaders.py
+++ b/kopf/utilities/loaders.py
@@ -14,11 +14,10 @@ Currently, two loading modes are supported, both are equivalent to Python CLI:
 Multiple files/modules can be specified. They will be loaded in the order.
 """
 
-import sys
 import importlib
 import importlib.util
 import os.path
-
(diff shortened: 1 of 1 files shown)


kopf/utilities/loaders.py
```diff
@@ -25,6 +26,7 @@ def preload(paths, modules):
     """
 
     for path in paths:
+        sys.path.insert(0, os.path.abspath(os.path.dirname(path)))
```
dneuhaeuser-zalando: This might add duplicate paths to `sys.path`, if `paths` contains the paths to multiple files in the same directory. If no de-duplication is performed during importing this could slow down imports a bit.

nolar: Indeed. 

But only in the case when many scripts from the same dir are put on the command-line:

```
kopf run di/script1.py ... dir/script100.py
```

Which is hard to imagine. Even 2 is rarely needed (almost never) — as we learned while gaining some experience using Kopf.

So, let it be this way for now.

Issue #140, opened 2019-07-10
add --disable-certificate-verify command line argument


goern:
## Expected Behavior
Kopf does not validate the SSL certs, so that it will not fail with `[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1056)`

## Actual Behavior
SSL certs validation is enforce, and `kopf run` fails with `[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1056)` if run out of cluster.

## Steps to Reproduce the Problem

  1. `kopf run app.py --verbose --standalone`

## Specifications

  - Platform: PRETTY_NAME="Fedora 30 (Workstation Edition)"
  - Kubernetes version:
```
Client Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.4+7bd2e5b", GitCommit:"7bd2e5b", GitTreeState:"clean", BuildDate:"2019-05-19T23:52:43Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"11+", GitVersion:"v1.11.0+d4cacc0", GitCommit:"d4cacc0", GitTreeState:"clean", BuildDate:"2019-04-10T17:49:11Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"linux/amd64"}
```
  - Python version: Python 3.7.3

  - Python packages installed:
```
aiohttp==4.0.0a0
aiojobs==0.2.2
appdirs==1.4.3
async-timeout==3.0.1
attrs==19.1.0
black==19.3b0
cachetools==3.1.1
certifi==2019.6.16
chardet==3.0.4
Click==7.0
daiquiri==1.5.0
google-auth==1.6.3
idna==2.8
iso8601==0.1.12
kopf==0.19rc1
kubernetes==10.0.0
multidict==4.5.2
oauthlib==3.0.2
pip==19.1.1
pyasn1==0.4.5
pyasn1-modules==0.2.5
pykube-ng==0.25
python-dateutil==2.8.0
pytz==2019.1
PyYAML==5.1.1
requests==2.22.0
requests-oauthlib==1.2.0
rfc5424-logging-handler==1.4.3
rsa==4.0
sentry-sdk==0.10.1
setuptools==41.0.1
six==1.12.0
thoth-common==0.9.1
[...]

PR #141, opened 2019-07-10
Provide option to disable verification on login

fridex:
# One-line summary

Provide a new option to disable verification on login

## Description

Provide an option to explicitly turn off verification on login. This is especially useful if cluster certificates are not setup and the user is aware of turning off SSL certificate verification and all the implications turning off cert verification introduces.

## Types of Changes

- Configuration change

## Tasks
_List of tasks you will do to complete the PR_
  - [ ] Keep PR in a good shape to get it in :)

## Review
- [ ] Tests

Fixes: #140


nolar:
@fridex Sorry, I didn't get it from the PR's description. Can you please clarify how this change is related to SSL cert verification?

The `verify` parameter for `login()` call is used to verify the login: i.e. to make a sample API call to see if the configs/envvars are parsed properly, the token is present, and both TCP connection and HTTP(S) authentication do work (this indeed includes the SSL cert verification implicitly). 

But SSL cert verification in general is not affected by this `verify` parameter.

By disabling this auth verification, the login will pass normally. But the operator will fail on the next call to `GET ...`, `GET ...?watch=true`, and on any other API call. Or do I miss something?

fridex:
> @fridex Sorry, I didn't get it from the PR's description. Can you please clarify how this change is related to SSL cert verification?
> 
> The `verify` parameter for `login()` call is used to verify the login: i.e. to make a sample API call to see if the configs/envvars are parsed properly, the token is present, and both TCP connection and HTTP(S) authentication do work (this indeed includes the SSL cert verification implicitly).
> 
> But SSL cert verification in general is not affected by this `verify` parameter.
> 
> By disabling this auth verification, the login will pass normally. But the operator will fail on the next call to `GET ...`, `GET ...?watch=true`, and on any other API call. Or do I miss something?

@nolar, thanks for the review!

I didn't get implementation actually working on the cluster where I'm testing the code; I ended up with permissions denied on CR (I need to install it).

Looking at the implementation, the patch will not be that straightforward. I checked the verification part based on your comment, the code uses pykube and kubernetes package. 

Enforcing:

```python
    _pykube_cfg.cluster["insecure-skip-tls-verify"] = True
```

In `get_pykube_cfg()` does the trick for me and requests when pykube is used are insecure (I can see related urllib3 warnings and code passes).

However, I'm missing why there is a need to call `verify_client()` in `auth.py` which uses the official kubernetes python package as all the routines (hope I checked them all) use pykube. I found just the following comment in auth.py sources:

https://github.com/zalando-incubator/kopf/blob/700c7d14cd6d83c80f8a0d75bdea6f1dd37c5d14/kopf/clients/auth.py#L40-L41

Based on kubernets-python implementation, the instantiation of API client will need to change to respect modified configuration with `verify_ssl` set to `False`. Not sure how that affects other parts of sources, why was this verification implied?

goern:
fixes https://github.com/zalando-incubator/kopf/issues/140

nolar:
Thanks for you contribution. 

The whole authentication was fully reworked in #226. It now supports multiple credentials and dynamic on-demand re-authentication. 

And the verification is also fully gone — it is now implicitly done by trying the API calls, failing with 401, and triggering the re-authentication inside of the operator life time.

After the changes, I see no reasons to add a global CLI option to disable the SSL verification. First of all, it will be applied to all credentials, even when verification (or absence of it) is expected only on one. Second, it conceptually assumes more and more authentication flags to the Kopf's CLI, while K8s authentication is not the goal of this framework and should not be a big part of the user-facing features.

---

To achieve the desired goal of ignoring the SSL certificate validity, please use `@kopf.on.login()` handlers to produce a `kopf.ConnectionInfo(..., verify=False, ...)` instance.

To make it more simple, [`kopf.login_via_pykube()` and `kopf.login_via_client()`](https://kopf.readthedocs.io/en/latest/authentication/#piggybacking) can be used inside of the handler instead of own config parsing, of which the returned [`ConnectionInfo`](https://kopf.readthedocs.io/en/latest/packages/kopf/#kopf.ConnectionInfo) object can be modified with [`dataclasses.replace(result, verify=False)`](https://docs.python.org/3/library/dataclasses.html#dataclasses.replace).

Few examples can be found here:

* Docs: https://kopf.readthedocs.io/en/latest/authentication/
* Example: https://github.com/zalando-incubator/kopf/blob/master/examples/13-hooks/example.py

Please report any issues with the new custom authentication methods if they do not work as expected.

nolar:
PS: Pre-released as `kopf==0.23rc1`. Not yet a stable release (coming in few days).

fridex:
> Thanks for you contribution.
> 
> The whole authentication was fully reworked in #226. It now supports multiple credentials and dynamic on-demand re-authentication.
> 
> And the verification is also fully gone — it is now implicitly done by trying the API calls, failing with 401, and triggering the re-authentication inside of the operator life time.
> 
> After the changes, I see no reasons to add a global CLI option to disable the SSL verification. First of all, it will be applied to all credentials, even when verification (or absence of it) is expected only on one. Second, it conceptually assumes more and more authentication flags to the Kopf's CLI, while K8s authentication is not the goal of this framework and should not be a big part of the user-facing features.
> 
> To achieve the desired goal of ignoring the SSL certificate validity, please use `@kopf.on.login()` handlers to produce a `kopf.ConnectionInfo(..., verify=False, ...)` instance.
> 
> To make it more simple, [`kopf.login_via_pykube()` and `kopf.login_via_client()`](https://kopf.readthedocs.io/en/latest/authentication/#piggybacking) can be used inside of the handler instead of own config parsing, of which the returned [`ConnectionInfo`](https://kopf.readthedocs.io/en/latest/packages/kopf/#kopf.ConnectionInfo) object can be modified with [`dataclasses.replace(result, verify=False)`](https://docs.python.org/3/library/dataclasses.html#dataclasses.replace).
> 
> Few examples can be found here:
> 
> * Docs: https://kopf.readthedocs.io/en/latest/authentication/
> * Example: https://github.com/zalando-incubator/kopf/blob/master/examples/13-hooks/example.py
> 
> Please report any issues with the new custom authentication methods if they do not work as expected.

@nolar thanks for the info. The described design looks neat! :+1:

Issue #142, opened 2019-07-10
Slow or hanging process termination


nolar:
## Expected Behavior

When an exception happens, or SIGTERM/SIGINT received, the operator must exit immediately or within a short time (e.g. 1s).

## Actual Behavior

In PyCharm, when the process is stopped, it can hang there for long time, up to few minutes. If `default_stream_timeout` is set, the wait time is usually limited to at most that many seconds.

In a terminal, clicking Ctrl+C does not kill the process, and requires the second Ctrl+C click.

In a pod, the operator can freeze in a semi-dead state, where it does not handle the events, but also does not exit normally and prevents the pod restart.

## Specifications

  - Version: 0.19rc1, 0.18

nolar:
Finally solved with `kopf==0.23rc1` ([release notes](https://github.com/zalando-incubator/kopf/releases/tag/0.23rc1)) and full switch to aiohttp (#227). The operator now terminates almost instantly.

PR #143, opened 2019-07-10
Die properly: fast and with no remnants

nolar:
When an operator gets SIGINT (Ctrl+C) or SIGTERM (in pods), it exits immediately, with no extra waits for anything.

This also improves its behaviour in PyCharm, where the operator can now be stopped and restarted in one click (previously required double-stopping).

> Issue : #142 
> Requires: https://github.com/hjacobs/pykube/pull/31

### Watch-stream termination

There is no easy way to terminate the blocking `read()` operation on a socket. The only way is a dirty hack with the OS-level process signals, which interrupt the I/O operations on low (libc&co) level. 

This hack can be done only from the main thread (because signal handlers can be set only from the main thread — a Python's limitation), and therefore is only suitable for the runnable application rather than for a library.

In `kopf run`, we can be sure that the event-loop and its asyncio tasks run in the main thread. In case of explicit task orchestration, we can assume that the operator developers do the same, but warn if they don't. 

This PR:

* Moves the streaming watch-requests into their dedicated threads (instead of the asyncio thread-pool executors for every `next()` call).
* These dedicated threads are tracked in each `watch_objs()` call, and are interrupted when the watch-stream exits for any reason (an error, generator-exit, task cancellation, etc).

### Tasks cancellation

Beside of the stream termination, this PR improves the way how the tasks are orchestrated in an event-loop, so that they are cancelled properly, and have some minimal time for graceful exit (e.g. cleanups). This is used, for example, by the peering `disappear()` call to remove self from the peering object (previously, this call was cancelled on start, and therefore never actually executed).

Also, the sub-tasks of the root tasks are tracked and also cancelled. For example, the scheduled workers, handlers, all shielded tasks, and generally anything that is not produced by `create_tasks()`, but runs in the loop. This guarantees, that when `run()` exits, it leaves nothing behind.

## Types of Changes

- Refactor/improvements

## Review
- [ ] Tests
- [ ] Documentation


nolar:
Split in favour of separate #147 #148 #149 #152 — one per fix.

Issue #136, opened 2019-07-04
K8s-event logging


nolar:
It would be convenient, if all the important log messages are automatically sent to K8s events, same as they are logged to stdout/stderr.

Only the important messages, not the debug messages.

Both Kopf-internal messages (already sent to k8s now), and the operator-produced messages (as logged via the per-object `logger` kwarg).

There should be a way to opt out at least on a per-message level — i.e. to explicitly say to not send it as a k8s-event, but to keep it as a log message only (except as logging it as a debug-message).

```python
import kopf

@kopf.on.create('zalando.org', 'v1', 'kopfexamples')
def create_fn(spec, logger, **kwargs):
    image = spec.get('image')
    if not image.startswith('trusted.repo.com/'):
        logger.warning("Only images from trusted.repo.com are trusted. "
                       "Got potentially insecure image %r", image)
```

This will hide Kopf's explicit k8s-event-sending routines — as it goes against Kopf's mission (of not being a K8s client library). But this will integrate Python `logging` machinery into K8s's logging/eventing subsystem — which perfectly fits into Kopf's mission of bringing Pythonic ways of coding to the K8s operators field.

PR #148, opened 2019-07-14
Post k8s-events properly from multiple threads (thread-safe)

nolar:
_Part of an effort to make Kopf-based operators more stable and resilient._

> Issue : #142 #136 
> Related: #128 

## Description

When k8s-event-posting was moved to a background task (#125), some asyncio-related tasks were made way to complicated (multiple dynamically created loggers) and implemented improperly (non-thread-safe `put_nowait()` when used across threads).

The non-thread-safeness made some k8s-events not posted as soon as it was possible, but waiting for few seconds until something happens on an asyncio event-loop (because `put_nowait()` does not wake up the event-loop). And it was failing of debug mode was used.

This has become important since #128 — as not only Kopf's internal routines, but also the sync-handlers (which run in a thread-pool) could log INFO+ messages, which should be posted to k8s.

This PR simplifies the solution to one and only one logger (and therefore one and only one handler), and by posting all k8s-events synchronously but instantly via one and only one function (`enqueue()`).

In that function (`enqueue()`), one of the thread-safe methods is used to put an event to a queue: either the same-threaded method (thread-safe by definition), or cross-threaded method (explicitly thread-safe).

## Types of Changes

- Refactor/improvements


kopf/engines/posting.py
```diff
@@ -38,12 +42,35 @@ class K8sEvent(NamedTuple):
...
+
+    if running_loop is loop:
+        # Posting from the same event-loop as the poster task and queue are in.
+        # Therefore, it is the same thread, and all calls here are thread-safe.
+        # Special thread-safe cross-event-loop methods make no effect here.
+        queue.put_nowait(event)
+    else:
+        # No event-loop or another event-loop - assume another thread.
+        # Use the cross-thread thread-safe methods. Block until enqueued there.
+        fut = asyncio.run_coroutine_threadsafe(queue.put(event), loop=loop)
```
samurang87: fut?

nolar: Some official docs use that too. E.g. [asyncio](https://docs.python.org/3/library/asyncio-future.html).

But right, better to have normal names. Fixed.

PR #152, opened 2019-07-16
Interrupt blocking socket reads on exit or error

nolar:
When an operator gets SIGINT (Ctrl+C) or SIGTERM (in pods), it exits immediately, with no extra waits for anything, and no double-killing needed.

_Part of an effort to make Kopf-based operators more stable and resilient._

> Issue: #142, partially replaces #143.

## Description

There is no easy way to terminate the blocking `read()` operation on a socket. The only way is a dirty hack with the OS-level process signals, which interrupt the I/O operations on low level (OS&libc&co). 

This hack can be done only from the main thread (because signal handlers can be set only from the main thread — a limitation of Python), and therefore is only suitable for the runnable application rather than for a library.

In `kopf run`, we can be sure that the event-loop and its asyncio tasks run in the main thread. In `KopfRunner` we can expect the test to be at least initiated from the main thread (but not rely on it). In case of explicit task orchestration, we can assume that the operator developers do the same, but warn if they don't.

This PR:

* Moves the streaming watch-requests into their dedicated threads (instead of the asyncio thread-pool executors for every `next()` call).
* These dedicated threads are tracked in each `watch_objs()` call, and are interrupted when the watch-stream exits for any reason (an error, generator-exit, task cancellation, etc).

## Types of Changes

- Refactor/improvements


kopf/clients/watching.py
```diff
@@ -83,51 +83,237 @@ def streaming_next(src):
...
+__interruption_lock = threading.Lock()
+__interruption_counter: int = 0
+__interruption_signal: Optional[int] = None
+__original_handler: Optional[Callable] = None
+
+
+def __can_interrupt():
+    return __interruption_signal is not None
+
+
```
nolar: The number of underscores emphasises how dark this magic is, and how much hidden it should remain.

nolar:
Closed in favour of #226 (auth) + #227 (aiohttp) — the problem is now solved by fully async i/o used in Kopf's API wrappers. No synchronous socket reads — no problem.

PR #156, opened 2019-07-24
Refactor runtime to full async/await; make officially embeddable

nolar:
Make operator termination even more graceful (resource-cleanable), the task orchestration — async, and the operator core — embeddable (and officially documented).

> Issue : #142.

## Description

Previously in #147, there was an attempt to rework the asyncio task management with the "should-stop" event and asyncio task cancellations. It made the operator to react to the existing signals, but still had some problems with proper cleanups.

Specifically, when a `aiohttp.ClientSession` is added (a PR based on this one — to be created), there is no place to properly create it and close it, since all the tasks are orchestrated in a synchronous function.

This led to few issues with some `aiohttp` connections not being closed properly, issuing the warnings/prints on the process exiting (and a lot of them in tests).

### Fully async coroutines

With this PR, and operator's body is extracted into a self-contained coroutine (`operator()`), which does everything: task spawning, task orchestration, resource creation and cleanups, etc.

The synchronous functions `run()` and `create_tasks()` are left as wrappers for `operator()` and `spawn_tasks()` respectively. (`create_tasks()` is also deprecated.)

### Graceful termination

With this PR, the operator termination is now even more graceful both in case of normal exit ("should-stop" event set) and in case of error. In the latter case, all tasks & sub-tasks are cancelled without a tolerable graceful timeout, but they are still given time to cleanup their resources (e.g. in `try-finally`).

## Embeddability

These function were already exposed via the public interface (accessible via `kopf.run`/`kopf.create_tasks`). 

With this PR, all the new functions are also exposed, and this is made a documented feature of embeddability — i.e. ability to embed Kopf operators into third-party applications by explicit invocation/orchestration of the Kopf tasks.

_(As an example of potential usage (not in the plans): an application that listens for the resources updates and keeps their state in-memory, while showing that state in a UI/WebUI.)_ 

## Types of Changes

- Refactor/improvements
- Documentation / non-code

## Review
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/docs/embedding.rst
+++ b/docs/embedding.rst
@@ -105,8 +105,8 @@ in :mod:`contextvars` containers with values isolated per-loop and per-task.
 .. warning::
     It is not recommended to run Kopf in the same event-loop with other routines
     or applications: it considers all tasks in the event-loop as spawned by its
-    workers and handlers, and cancells them when it exits.
+    workers and handlers, and cancels them when it exits.
 
-    There are some basic safety measures to not cancel the tasks existed before
-    operator startup, but that cannot be applied to the tasks spawned later
-    due to asyncio implementation details.
+    There are some basic safety measures to not cancel tasks existing prior
+    to the operator's startup, but that cannot be applied to the tasks spawned
+    later due to asyncio implementation details.
--- a/docs/embedding.rst
+++ b/docs/embedding.rst
@@ -1,0 +1,112 @@
+=========
+Embedding
+=========
+
+Kopf is designed to be embeddable into other applications, which require
+watching over the Kubernetes resources (custom or built-in), and handling
+the changes.
+This can be used, for example, in desktop applications or web APIs/UIs
+to keep the state of the cluster and its resources in memory.
+
+
+Manual orchestration
+====================
+
+Since Kopf is fully asynchronous, the best way to run Kopf is to provide
+an event-loop specially for Kopf in a separate thread, while running
+the main application in the main thread.
+
+.. code-block:: python
+
+    import asyncio
+    import threading
+
+    import kopf
+
+    @kopf.on.create('zalando.org', 'v1', 'kopfexamples')
+    def create_fn(**_):
+        pass
+
+    def kopf_thread():
+        loop = asyncio.get_event_loop()
+        loop.run_until_complete(kopf.operator())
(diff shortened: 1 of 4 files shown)


docs/embedding.rst
```diff
@@ -0,0 +1,112 @@
...
+        thread = threading.Thread(target_fn=kopf_thread)
+        thread.start()
+        # ...
+        thread.join()
+
+
+.. warning::
+    It is not recommended to run Kopf in the same event-loop with other routines
+    or applications: it considers all tasks in the event-loop as spawned by its
+    workers and handlers, and cancells them when it exits.
```
samurang87: "cancels"

docs/embedding.rst
```diff
@@ -0,0 +1,112 @@
...
+
+    There are some basic safety measures to not cancel the tasks existed before
```
samurang87: "cancel tasks existing prior to the operator's startup"

kopf/reactor/running.py
```diff
@@ -130,29 +149,104 @@ def create_tasks(
...
+            # If the operator is cancelled, propagate the cancellation to all the sub-tasks.
+            hung_cancelled, hung_left = await _stop(hung_tasks, title="Hung", cancelled=True)
+            raise
+        else:
+            # If the operator is intact, but the timeout is reached, forcely cancel the sub-tasks.
+            hung_cancelled, hung_left = await _stop(hung_pending, title="Hung", cancelled=False)
+
+    # If succeeded or if cancellation is silenced, re-raise from failed tasks (if any).
+    await _reraise(root_done | root_cancelled | hung_done | hung_cancelled)
+    # assert not root_left
```
samurang87: are the comments intentional? :)

docs/embedding.rst
nolar: Fixed.

docs/embedding.rst
nolar: Fixed.

kopf/reactor/running.py
nolar: Removed.

nolar:
@samurang87 Thanks. All is fixed now.

psycho-ir:
👍

Issue #139, opened 2019-07-10
Error on already removed object patch


dbazhal:
## Expected Behavior
When I create and quickly remove object, that's an odd behavior, but I expect that kopf will handle it as non-error.

## Actual Behavior
Insetad, I get errors in log and in error tracker:
```
ApiException: (404)
Reason: Not Found
HTTP response headers: HTTPHeaderDict({'Cache-Control': 'no-store', 'Content-Type': 'application/json', 'Date': 'Wed, 10 Jul 2019 10:03:20 GMT', 'Content-Length': '222'})
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"increments.tutu.ru \"inc-e5c1\" not found","reason":"NotFound","details":{"name":"inc-e5c1","group":"tutu.ru","kind":"increments"},"code":404}


  File "kopf/reactor/queueing.py", line 145, in worker
    await handler(event=event)
  File "kopf/reactor/handling.py", line 139, in custom_object_handler
    await patching.patch_obj(resource=resource, patch=patch, body=body)
  File "kopf/clients/patching.py", line 41, in patch_obj
    await loop.run_in_executor(config.WorkersConfig.get_syn_executor(), functools.partial(patch_func, **request_kwargs))
  File "concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "kubernetes/client/apis/custom_objects_api.py", line 1951, in patch_namespaced_custom_object
    (data) = self.patch_namespaced_custom_object_with_http_info(group, version, namespace, plural, name, body, **kwargs)
  File "kubernetes/client/apis/custom_objects_api.py", line 2057, in patch_namespaced_custom_object_with_http_info
    collection_formats=collection_formats)
  File "kubernetes/client/api_client.py", line 334, in call_api
    _return_http_data_only, collection_formats, _preload_content, _request_timeout)
  File "kubernetes/client/api_client.py", line 168, in __call_api
    _request_timeout=_request_timeout)
  File "kubernetes/client/api_client.py", line 393, in request
    body=body)
  File "kubernetes/client/rest.py", line 286, in PATCH
    body=body)
  File "kubernetes/client/rest.py", line 222, in request
    raise ApiException(http_resp=r)
```
I'm not sure what behavior will be correct in such situation, but maybe passing such situations will be ok?
```
except ApiException as e:
    if e.reason != 'NotFound':
        raise
```

[...]

Issue #24, opened 2019-04-06
Configurable/optional finalizers


nolar:
## Actual Behavior

[Finalizers](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/) are marks that are stored as a list of arbitrary strings in `metadata.finalizers`. When the object is requested to be deleted, its deletion timestamp is set only. Until such marks exist on the object, it will not be deleted, and the deletion request will wait until the marks are removed by the operators/controllers (which should react to the deletion timestamp appearance). Only when all the finalizers are removed, the object is actually deleted. 

Currently, when a resource is handled by the operator, the finalizers are always added on the object on its first appearance (before the handlers), and remove when it is marked for deletion (after the handlers).

If such an operator is stopped, the objects cannot be deleted, and the deletion command freezes — while there is no actual need to wait and to notify the operator (it will do nothing).

## Expected Behavior

The finalisers should be optional.  If there are no deletion handlers, the finalizers are not needed. If there are deletion handlers, the finalizers should be added.

Some deletion handlers can be explicitly marked as optional, thus ignored for the decision whether the finalizers are needed. The default assumption is that if the deletion handler exists, it is required.

```python
@kopf.on.delete('', 'v1', 'pods', optional=True)
def pod_deleted(**_):
    pass
```

Two special cases: 

* If the object was created when there were no deletion handlers, and the finalizers were not added, but then a new operator version is started with the deletion handlers — the finalizers must be auto-added.
* If the object was created when there were some deletion handlers, and the finalizers were added, but then a new operator version is started with no deletion handlers — the finalizers must be auto-removed.

dlmiddlecote:
Hey @nolar 

I’d like to have a go at this issue, I should have some time over the coming week. 

I have a preliminary question, if that’s ok.
In `kopf.reactor.causation.detect_cause` the finalizers are [used to check whether the event is `NEW` or not](https://github.com/zalando-incubator/kopf/blob/master/kopf/reactor/causation.py#L99-L105). I think by not adding the finalizers sometimes we will break that check, and so we’ll probably need some other way to do the check. 

Do you have any ideas what we could do? Or do you think things will still work fine?

nolar:
@dlmiddlecote Okay, I will find a way to assign it to you (it didn't work that easily as I thought).

[...]

PR #160, opened 2019-08-01
Ignore "404 Not Found" when patching after deletion during handling

nolar:
Ignore errors when patching an object that was removed too quickly, during the handling cycle or inside of the handlers.

> Issue : #139

## Description

If a handler deletes the object being handled, Kopf will try to patch its status after the handler as usually, and will fail, since the object does not exist anymore.

The same happens if the object is deleted too quickly, while the handler is running. 

Previously, this was prevented by always having a finalizer before any handling. Since #24, the finalizer is added only when there are non-optional delete handlers, and skipped otherwise. So, it is now possible that the object is deleted while we handling it.

## Types of Changes

## Review

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/kopf/clients/patching.py
+++ b/kopf/clients/patching.py
@@ -40,8 +40,13 @@ async def patch_obj(*, resource, patch, namespace=None, name=None, body=None):
         await loop.run_in_executor(config.WorkersConfig.get_syn_executor(), obj.patch, patch)
     except pykube.ObjectDoesNotExist:
         pass
+    except pykube.exceptions.HTTPError as e:
+        if e.code == 404:
+            pass
+        else:
+            raise
     except requests.exceptions.HTTPError as e:
-        if e.response.status_code in [404]:
+        if e.response.status_code == 404:
             pass
         else:
             raise
--- a/kopf/clients/patching.py
+++ b/kopf/clients/patching.py
@@ -1,5 +1,8 @@
 import asyncio
 
+import pykube
+import requests
+
 from kopf import config
 from kopf.clients import auth
 from kopf.clients import classes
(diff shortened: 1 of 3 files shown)


kopf/clients/patching.py
```diff
@@ -31,5 +34,14 @@
...
 
-    loop = asyncio.get_running_loop()
-    await loop.run_in_executor(config.WorkersConfig.get_syn_executor(), obj.patch, patch)
+    # The handler could delete its own object, so we have nothing to patch. It is okay, ignore.
+    try:
+        loop = asyncio.get_running_loop()
+        await loop.run_in_executor(config.WorkersConfig.get_syn_executor(), obj.patch, patch)
+    except pykube.ObjectDoesNotExist:
+        pass
+    except requests.exceptions.HTTPError as e:
```
psycho-ir: can you explain in what circumstances this exception might raise?

nolar: Indeed! Good that you've noticed that.

Pykube has [its own handling of _all_ exceptions](https://github.com/hjacobs/pykube/blob/2ff99db72be0207af4f0c6e2c6d82adbeeb63559/pykube/http.py#L242-L249), and overrides it with its own `pykube.HTTPError` _only_ if they are applicaton/json and its json has `kind=="Status"` — regardless of the status code. I don't know how reliable this condition is, so it might happen that the `requests`'s exceptions leak out.

But that pykube's exception should be caught too, same as in the `auth.py` (on verification).

And, unlike the GET-methods in `fetching.py`, `pykube.ObjectDoesNotExist` will not be raised in PATCH/POST/PUT-methods. But that might change in the future, so we at least expect it (for forward-compatibility).

nolar: Fixed with a new commit.

psycho-ir: Right, being more conservative in there regards shouldn't hurt

Issue #161, opened 2019-08-02
No immediate reaction on changes/deletion if sleeping for retries


nolar:
## Expected Behavior

When a handler function raises an exception, it is scheduled to be retried later (by default, in 60 seconds). Other handlers will be invoked — until the state is reached, when all the remaining handlers are scheduled for later.

When a handler with an exception or a retry is scheduled for later execution, and its operator is sleeping until the time comes, and a new watch-event arrives — the sleep must be interrupted, and a new state must be processes immediately, not after few tens of seconds.

## Actual Behavior

Kopf uses unconditional `asyncio.sleep()` with the shortest duration until the next handler to be awakened. 

If new events arrive for that object — e.g. it was changed or deleted — the sleep continues unconditionally, and the  new events are processes much later.

## Steps to Reproduce the Problem

```python
import kopf

@kopf.on.create('zalando.org', 'v1', 'kopfexamples')
def create(**_):
    raise Exception("oops")
```

  1. Run the operator in the background.
  1. Create an object.
  1. Within the next 60 seconds (default timeout), delete the object.
  1. Observe: the deletion is frozen until the operator reaches its 60s timing.

## Specifications

```
kopf==0.20
```

Issue #150, opened 2019-07-15
Reconciling resources


klarose:
## Expected Behavior
An important part of kubernetes is the idea of eventual consistency. If a document is applied to the cluster, the state it describes may not take effect immediately, or the cluster may deviate occasionally from it. I want to be able to write operators which take this principle to heart.

In order to do this, I think that we need to be able to add something akin to [reconciliation loops](https://github.com/operator-framework/operator-sdk/blob/master/doc/user-guide.md#reconcile-loop) in https://github.com/operator-framework/operator-sdk.

For example, if I create a resource, say `MyDeployment`. `MyDeployment` contains allows me to create k8s `Deployment` objects simplified for my specific Docker image.

If something changes the `Deployment` generated by `MyDeployment`, I want to be notified of that change so that I may place it back into the desired state.

## Actual Behavior
From reading the documentation, there doesn't appear to be any way to be notified when an 'owned' resource deviates from the desired state. It would be nice if we could do this in a similar decorator-style fashion as the other `kopf.on` module decorators. I haven't thought enough about what a good API would look like here. I think it would be necessarily different from the go operator-sdk method, since it handles everything in one loop, wheareas kopf splits things up for you..

## Other stuff

Thanks for doing this! I've implemented a few off-the-cuff operators in python. What you've done here would have simplified them dramatically.

nolar:
@klarose First of all, thanks for your warm words! This keep me motivated and drives the project :-) 

Also thanks for the detailed suggestion. This in indeed an important feature to have, and we discussed it from different points in other contexts too.

A very important part of a reconciliation loop is knowing the state of the children. In the current implementation, Kopf only knows the state of the operated resource. In some cases, the children can be other k8s resources, but it is also possible to have some third-party API children objects: orchestrate from K8s via CRDs, but execute/store in an external environment via APIs.

Kopf itself cannot manage all of such states. But it can help.

---

Currently, a reconciliation loop can be implemented via `@kopf.on.resume + @kopf.on.create` handlers (the same function with 2 decorators). The handler should start a thread or a task with `while True` and polling the state of the children/external resources every N seconds.

An example of this is in [examples/10/builtins](https://github.com/zalando-incubator/kopf/blob/master/examples/10-builtins/example.py). Also see #122.

```python
import asyncio
import kopf

@kopf.on.resume('zalando.org', 'v1', 'KopfExample')
@kopf.on.create('zalando.org', 'v1', 'KopfExample')
[...]

PR #163, opened 2019-08-02
Ensure that all handlers get the latest state (eventually)

nolar:
Keep track of which state was handled by which handler, and re-execute it if the state is changed during the handling cycle.

> Issue : replays #9, indirectly #161 and #150

## Description

In #9, there was an attempt to make the handler's status persistent. As a side-effect, it also could detect the state changes during the handling cycle. The persistent progress was not needed or not much useful. However, the change detection is needed again. This PR partially replays the same solution.

Currently, if an object is changed during the long handling cycle (which can include multiple handlers and multiple retries — i.e. for hours sometimes), these changes will not be noticed in some handlers. The "last-seen" state will be stored as it was at the moment of the _last_ handler, and only the changes relative to that state will be handled since then. 

**As a result,** if some handler happened to run before the change was introduced, i.e. it has got an old state, that handler would never know about the new state.

With this PR, **all update & deletion handlers** handlers will track the handled state of the objects individually (as a short hash-digest). If it changes during the handling cycle, those handlers will be re-executed with the new state. 

**This guarantees that every handler will see the latest actual state of the object — the closest to the concept of reconciliation.**

**For creation handlers,** the behaviour is slightly different. In case of changes during the creation cycle, the creation handlers will be continued as normally (not re-executed). And only after the full creation cycle is done, an update cycle will start with the changes relatively to the _first_ (not last!) creation handler, i.e. when the object was seen in the first time. 

This should prevent spawning duplicate children objects, which is usually done in the creation handlers.  The update handlers should be designed idempotent anyway, so it is safe to shift the repeated handling of the same state from the creation handlers to the update handlers.

## Types of Changes

## Review

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/kopf/reactor/causation.py
+++ b/kopf/reactor/causation.py
@@ -138,6 +138,7 @@ def detect_cause(
     # For an object seen for the first time (i.e. just-created), call the creation handlers,
     # then mark the state as if it was seen when the creation has finished.
     if not lastseen.has_state(body):
+        kwargs['digest'] = True  # or any other true'ish constant (str/int)
         return Cause(
             event=CREATE,
             body=body,
--- a/kopf/reactor/causation.py
+++ b/kopf/reactor/causation.py
@@ -106,6 +106,7 @@ def detect_cause(
             initial=initial,
             **kwargs)
 
+    # Marked for deletion, but we still hold it with our finalizer.
     if finalizers.is_deleted(body):
         return Cause(
             event=DELETE,
--- a/kopf/reactor/causation.py
+++ b/kopf/reactor/causation.py
@@ -67,6 +67,7 @@ class Cause(NamedTuple):
     initial: bool
     body: MutableMapping
     patch: MutableMapping
+    digest: Union[str, int]
     diff: Optional[diffs.Diff] = None
     old: Optional[Any] = None
     new: Optional[Any] = None
--- a/kopf/structs/lastseen.py
+++ b/kopf/structs/lastseen.py
@@ -14,6 +14,7 @@ https://kubernetes.io/docs/concepts/overview/object-management-kubectl/declarati
 """
 
 import copy
+import hashlib
 import json
 
 from kopf.structs import dicts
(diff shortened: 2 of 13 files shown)


nolar:
Rebased on master with amendments for the new codebase & features.

kopf/structs/lastseen.py
```diff
@@ -98,5 +99,29 @@ def retreive_state(body):
...
+
+def freeze_state(*, body, patch, extra_fields=None):
+    frozen_state = body.get('status', {}).get('kopf', {}).get('frozen-state')
+    actual_state = get_state(body, extra_fields=extra_fields)
+    if frozen_state is None:
+        storage = patch.setdefault('status', {}).setdefault('kopf', {})
+        storage['frozen-state'] = json.dumps(actual_state)
+
+
+def compute_digest(body, extra_fields=None):
```
psycho-ir:
```suggestion
def compute_digest(body, extra_fields=None) -> str:
```

kopf/reactor/causation.py
```diff
@@ -67,6 +67,7 @@ class Cause(NamedTuple):
     initial: bool
     body: MutableMapping
     patch: MutableMapping
+    digest: Union[str, int]
```
psycho-ir: in what conditions the `digest` can be `int`?

nolar: When it is reimplemented to something else (e.g. crc32). We just show that we do not care that it is necessary a string.

psycho-ir: I see, so in the specification we don't care wether this string or int.

kopf/structs/lastseen.py
nolar: Good idea. I will do this in a separate refactoring PR with all the types put to all the functions all around (not 50/50 like it is now).

psycho-ir: Fair enough.

Issue #164, opened 2019-08-05
kopf can't post events


olivier-mauras:
## Expected Behavior
Be able to post events either custom of defaults.

## Actual Behavior
``` text
Custom event with kopf.info()
[2019-08-05 15:09:56,422] kopf.clients.events  [WARNING ] Failed to post an event. Ignoring and continuing. Error: HTTPError('Event "kopf-event-tvlgd" is invalid: involvedObject.namespace: Invalid value: "": does not match event.namespace'). Event: type='Normal', reason='SA_DELETED', message='Managed service account got deleted'.

Default events:
[2019-08-05 15:12:47,217] kopf.clients.events  [WARNING ] Failed to post an event. Ignoring and continuing. Error: HTTPError('Event "kopf-event-5zplb" is invalid: involvedObject.namespace: Invalid value: "": does not match event.namespace'). Event: type='Normal', reason='Logging', message="Handler 'create_ns' succeeded.".
``` 

## Steps to Reproduce the Problem
No particular code is actually running
My controller is running with a cluster-admin role in it's own namespace.
``` yaml
---
apiVersion: v1
kind: ServiceAccount
metadata:
  namespace: kopf-controllers
  name: namespace-manager
---
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: ClusterRoleBinding
metadata:
  name: namespace-manager-rolebinding-cluster
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: cluster-admin
subjects:
  - kind: ServiceAccount
    name: namespace-manager
    namespace: kopf-controllers
```
A simple example like this gives the same result:
``` yaml
import kopf

@kopf.on.resume('', 'v1', 'namespaces')
@kopf.on.create('', 'v1', 'namespaces')
def create_fn(spec, **kwargs):
    print(f"And here we are! Creating: {spec}")
    return {'message': 'hello world'}  # will be the new status
```
Run with: `kopf run --standalone ./example.py `

``` text
...
[2019-08-05 15:19:54,389] kopf.clients.events  [WARNING ] Failed to post an event. Ignoring and continuing. Error: HTTPError('Event "kopf-event-27kj9" is invalid: involvedObject.namespace: Invalid value: "": does not match event.namespace'). Event: type='Normal', reason='Logging', message='All handlers succeeded for resuming.'.
[...]

PR #165, opened 2019-08-05
Post events for cluster-scoped objects to current namespace

nolar:
K8s-events cannot be posted cluster-scoped (?), and attached to cluster-scoped resources via API. However, we post them to the current namespace — so that they are not lost completely.

> Issue : #164

## Description

See #164 for details. 

Brief summary: K8s events are namespaced, there are no cluster-scoped events. Also, k8s events can refer to an object via spec.involvedObject of type ObjectReference. This structure contains namespace field, to refer to the involved object's namespace (also, name, uid, etc).

I could not find a way to post namespaced events for cluster-scoped resources. It always fails with namespace mismatch (regardless of which library is used). Event via `curl` — see the issue comments.

So, we post the k8s-events to the current namespace, so that they are available via `kubectl get events`, despite they are not seen in `kubectl describe` on the involved objects. 

It is not a full problem solution, but it is a bit better than just losing them completely.

## Types of Changes

### DIFF
--- a/kopf/structs/hierarchies.py
+++ b/kopf/structs/hierarchies.py
@@ -7,14 +7,18 @@ from kopf.structs import dicts
 def build_object_reference(body):
     """
     Construct an object reference for the events.
+
+    Keep in mind that some fields can be absent: e.g. ``namespace``
+    for cluster resources, or e.g. ``apiVersion`` for ``kind: Node``, etc.
     """
-    return dict(
-        apiVersion=body['apiVersion'],
-        kind=body['kind'],
-        name=body['metadata']['name'],
-        uid=body['metadata']['uid'],
-        namespace=body['metadata']['namespace'],
+    ref = dict(
+        apiVersion=body.get('apiVersion'),
+        kind=body.get('kind'),
+        name=body.get('metadata', {}).get('name'),
+        uid=body.get('metadata', {}).get('uid'),
+        namespace=body.get('metadata', {}).get('namespace'),
     )
+    return {key: val for key, val in ref.items() if val}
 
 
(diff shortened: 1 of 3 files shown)


kopf/structs/hierarchies.py
```diff
@@ -44,7 +52,7 @@ def append_owner_reference(objs, owner):
     owner = build_owner_reference(owner)
     for obj in dicts.walk(objs):
         refs = obj.setdefault('metadata', {}).setdefault('ownerReferences', [])
-        matching = [ref for ref in refs if ref['uid'] == owner['uid']]
+        matching = [ref for ref in refs if ref.get('uid') == owner.get('uid')]
```
psycho-ir: is this safe to keep refs with id None if the owner's uid is None? 
not sure if both owner's uid and ref's uid can be None is the same time though!

nolar: Perhaps, `ref['uid']` is mandatory. I was replacing these things in almost automated way, didn't pay attention (previously, it would fail with a KeyError in such cases).

The only case, when this can happen so that `ref['uid']` is None/absent, is when we create both the owner & the child, and keep them in-memory, not yet posted to the API, therefore not yet having the uids — and then we try to connect them. Adding an additional 2nd owner would hit this mach `None == None`, despite the 1st & 2nd owners can have different apiVersion/kind/name.

This is quite an artificial case. I'm not even sure if it is possible to add owner references without a uid, just by a name. We can solve it in advance, but maybe as a separate PR — not related to the event-posting. And I'm not even sure how to properly reproduce this scenario.

Issue #166, opened 2019-08-06
Invalid attribute apiVersion in build_object_reference()


olivier-mauras:
## Actual Behavior
``` text
[2019-08-06 05:50:32,627] kopf.objects         [ERROR   ] [xxx-dev/tiller] Handler 'sa_delete' failed with an exception. Will retry.
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/kopf/reactor/handling.py", line 387, in _execute
    lifecycle=lifecycle,  # just a default for the sub-handlers, not used directly.
  File "/usr/local/lib/python3.7/site-packages/kopf/reactor/handling.py", line 478, in _call_handler
    **kwargs,
  File "/usr/local/lib/python3.7/site-packages/kopf/reactor/invocation.py", line 66, in invoke
    result = await fn(*args, **kwargs)
  File "./ns.py", line 235, in sa_delete
    kopf.info(ns.to_dict(), reason='SA_DELETED', message='Managed service account got deleted')
  File "/usr/local/lib/python3.7/site-packages/kopf/engines/posting.py", line 79, in info
    event(obj, type='Normal', reason=reason, message=message)
  File "/usr/local/lib/python3.7/site-packages/kopf/engines/posting.py", line 72, in event
    ref = hierarchies.build_object_reference(obj)
  File "/usr/local/lib/python3.7/site-packages/kopf/structs/hierarchies.py", line 12, in build_object_reference
    apiVersion=body['apiVersion'],
KeyError: 'apiVersion'
``` 

## Steps to Reproduce the Problem
``` python
@kopf.on.delete('', 'v1', 'serviceaccounts', annotations={'custom/created_by': 'namespace-manager'})
async def sa_delete(body, namespace,  **kwargs):
  try:
    api = kubernetes.client.CoreV1Api()
    ns = api.read_namespace(name=namespace)
  except ApiException as err:
    sprint('ERROR', 'Exception when calling CoreV1Api->read_namespace: {}'.format(err))

  kopf.info(ns.to_dict(), reason='SA_DELETED', message='Managed service account got deleted')
  return
```
#######
Culprit lies here in `hierarchies.py`
``` python
def build_object_reference(body):
    """
    Construct an object reference for the events.
    """
    return dict(
        apiVersion=body['apiVersion'],
        kind=body['kind'],
        name=body['metadata']['name'],
        uid=body['metadata']['uid'],
        namespace=body['metadata']['namespace'],
    )
```
[...]

PR #168, opened 2019-08-06
kopf/structs/hierarchies.py: handle apiVersion attribute inconsistene across api resources

olivier-mauras:
# One-line summary
handle apiVersion attribute inconsistene across api resources
> Issue : #166 

## Description
The attribute apiVersion can be api_version depending of api resource and objects.

## Types of Changes

### DIFF
--- a/kopf/structs/hierarchies.py
+++ b/kopf/structs/hierarchies.py
@@ -4,12 +4,27 @@ All the functions to properly build the object hierarchies.
 from kopf.structs import dicts
 
 
+def get_api_version(body):
+    """
+    Api version attribute may differ across api resources
+    Returns the correct one
+    """
+    if 'apiVersion' in body:
+        api_ver = body['apiVersion']
+    elif 'api_version' in body:
+        api_ver = body['api_version']
+
+    return api_ver
+
+
 def build_object_reference(body):
     """
     Construct an object reference for the events.
     """
+    api_version = get_api_version(body)
+
     return dict(
-        apiVersion=body['apiVersion'],
+        apiVersion=api_version,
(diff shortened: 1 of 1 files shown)


kopf/structs/hierarchies.py
```diff
@@ -8,8 +8,14 @@ def build_object_reference(body):
     """
     Construct an object reference for the events.
     """
+    # apiVersion attribute is not consistent across api resources
+    if 'apiVersion' in body:
```
psycho-ir: Thank you so much for contributing to this project. 
Would be great to extract this code to a function to reduce the duplication.

olivier-mauras: Done

psycho-ir: Thanks @olivier-mauras, 
Would also be great if you could add one or more resource names as example for each attribute name.

olivier-mauras: I added a  comment https://github.com/zalando-incubator/kopf/issues/166#issuecomment-519126078 on the bug I opened.
It's actually not the API resources that differ but the body returned by on.xxxx() handlers.

nolar:
@olivier-mauras Thanks for your reported issue and contribution. Hopefully, there will be more in the future.

_For historic reference:_ The reason why this PR and this fix are not accepted is explained in https://github.com/zalando-incubator/kopf/issues/166#issuecomment-519171462. **TL;DR:** It is the Kubertenes Client's specific convention of snake-cased `api_version`; the K8s API provides and expects the camel-cased `apiVersion` (same is true for other multi-word fields).

Issue #167, opened 2019-08-06
Annotations/label filtering on on.delete() handler doesn't seem to work


olivier-mauras:
## Expected Behavior
With an on.delete() handler filtering on annotations, I expected to retrieve the kopf finalizer only on filtered resources.

## Actual Behavior
All resources handled by the on.delete() handler end up with the kopf finalizer making me believe the filtering either by annotation or label to not work correctly

## Steps to Reproduce the Problem
``` python
@kopf.on.delete('', 'v1', 'serviceaccounts', annotations={'custom/created_by': 'namespace-manager'})
async def sa_delete(body, namespace,  **kwargs):
  pass
```

Result:
``` text
$ kubectl -n trident get sa tiller -o yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  annotations:
    kopf.zalando.org/last-handled-configuration: '{"secrets": [{"name": "tiller-token-mxn8p"}],
      "metadata": {"labels": {"custom/name":      "trident", "custom/team": "trident"}, "annotations": {"custom/created_by":
      "namespace-manager"}}}'
    custom/created_by: namespace-manager
  creationTimestamp: 2019-08-06T06:17:51Z
  finalizers:
  - kopf.zalando.org/KopfFinalizerMarker
  labels:
    custom/name: trident
    custom/team: trident
  name: tiller
  namespace: trident
secrets:
- name: tiller-token-mxn8p

$ kubectl -n trident get sa default -o yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  annotations:
    kopf.zalando.org/last-handled-configuration: '{"secrets": [{"name": "default-token-bzjqv"}]}'
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","kind":"ServiceAccount","metadata":{"annotations":{"kopf.zalando.org/last-handled-configuration":"{\"secrets\": [{\"name\": \"default-token-bzjqv\"}]}"},"creationTimestamp":"2019-02-04T12:18:54Z","name":"default","namespace":"trident","resourceVersion":"27507246","selfLink":"/api/v1/namespaces/trident/serviceaccounts/default","uid":"0a321e2f-2877-11e9-a525-005056abd413"},"secrets":[{"name":"default-token-bzjqv"}]}
  creationTimestamp: 2019-02-04T12:18:54Z
  finalizers:
  - kopf.zalando.org/KopfFinalizerMarker
  name: default
  namespace: trident
secrets:
[...]

PR #171, opened 2019-08-06
Don't add finalizer to skipped objects

dlmiddlecote:
> Issue : #167

## Description
Don't add finalizer to object if there are no handlers for it. Now that it is possible to filter objects out of handler execution, this is pertinent. 

## Types of Changes

## Tasks
  - [ ] Add Tests

## Review
- [ ] Tests
- [ ] Documentation


psycho-ir:
Hey @dlmiddlecote, 

Thank you so much for the contribution. I just did manual test this change set and managed to reproduce #167 again. 

Could you please do a manual test if you haven't done yet?

kopf/reactor/handling.py
```diff
@@ -240,8 +240,8 @@ class HandlerChildrenRetry(HandlerRetryError):
     # finalizers, lock it to this operator. Not all newly created objects will
     # produce an 'ACQUIRE' causation event. This only happens when there are
     # mandatory deletion handlers registered for the given object, i.e. if finalizers
-    # are required.
-    if cause.event == causation.ACQUIRE:
+    # are required. Only do this if there are handlers to execute.
+    if not skip and cause.event == causation.ACQUIRE:
```
psycho-ir: I think this condition cannot be True as for not `HANDLER_CAUSES` the `skip` variable is aways None.

dlmiddlecote:
Hey!

Of course, no problem. I thought I had fixed this as I didn’t observe the same behaviour when I tried this, but I’ll definitely double check!

dlmiddlecote:
Yes, I’ve completely been mistaken here. I’ll look at this again. Sorry for the bother.

dlmiddlecote:
Closing this and starting again.

kopf/reactor/handling.py
nolar: True. It could be `if handlers and ...` here.

Keep in mind that if a finalizer is not added because of this `if` here, the `detect_cause()` will believe that it should be added on the next iteration, and so on forever, until it is actually added. The cause handling will not happen until then.

The best solution here would be to **_not_** detect the `ACQUIRE` cause if there are no deletion handlers for that specific individual object.

Now, it is done above in the same file by a resource kind only: `requires_finalizer=registry.requires_finalizer(resource=resource)`. It should be the whole individual object, it seems, not just a resource kind. 

---

Keep in mind a tricky scenario:

* A non-matching object is created.
* No deletion handlers are selected by filters.
* Therefore, no finalizer is added.
* The handling cycle is done, the system is idling.
* Suddenly, the object gets its labels/annotations assigned, which makes it matching the filters.

Before starting an `UPDATE` cause handling, the `ACQUIRE` cause should be pre-executed, and the finalizer should be added (because the criteria for its existence have changed).

nolar:
@dlmiddlecote Oh, sorry, I didn't notice the PR is closed. Please, see my comment above with some hints.

PR #173, opened 2019-08-07
Don’t add finalizers to skipped objects

dlmiddlecote:
> Issue : #167

## Description
Don't add finalizer to object if there are no handlers for it. Now that it is possible to filter objects out of handler execution, this is pertinent. 

## Types of Changes

## Tasks
  - [x] Add Tests

## Review
- [ ] Tests
- [ ] Documentation


dlmiddlecote:
Tests seem to be failing on 1 flakey test.

psycho-ir:
Hi @dlmiddlecote, 

Thank you so much for the PR! 

@nolar 
I tested it locally and it works fine in almost all the cases I had in mind. 
The only scenario that it doesn't work correctly is when we `annotate` a resource and it becomes matched with on of the registered handlers, the finalizer won't be added to the resource.

dlmiddlecote:
Hey @psycho-ir 

Is this the case?
- resource with no annotations applied => no finalizer applied
- resource edited (or `kubectl annotate` used) to add matching annotations => finalizer should be applied?

If so, I just tried this, and it seems to work.

Let me know if I'm mistaken.

tests/registries/test_requires_finalizer.py
```diff
@@ -0,0 +1,152 @@
...
+def test_requires_finalizer_deletion_handler(optional):
+    registry = GlobalRegistry()
+    resource = Resource('group', 'version', 'plural')
+
+    @kopf.on.delete('group', 'version', 'plural',
+                    registry=registry, optional=optional)
+    def fn(**_):
+        pass
+
+    assert registry.requires_finalizer(resource=resource, body=OBJECT_BODY) is not optional
```
nolar: This construct is a bit counter-intuitive. Maybe, parametrize with 2 values: `@pytest.mark.parametrize('optional, expected'...`, and then compare with `== expected`.

Also, it is better to put the actual result of a function call to a variable, and then use it in the assertion. Pytest then renders the failed assertions nice (and not so nice in some case when the function call is used: because double-calling for rendering can have side-effects).

psycho-ir:
> Hey @psycho-ir
> 
> Is this the case?
> 
> * resource with no annotations applied => no finalizer applied
> * resource edited (or `kubectl annotate` used) to add matching annotations => finalizer should be applied?
> 
> If so, I just tried this, and it seems to work.
> 
> Let me know if I'm mistaken.

Hi @dlmiddlecote, 

Sorry you are right this scenario works perfectly fine.
What didn't work as expected for me was the other way around:

* resource with annotation applied => finalizer applied
* resource patched (annotation removed) => finalizer is still there

dlmiddlecote:
Hey!

I also can't reproduce.

I have the operator as:
```
import kopf

@kopf.on.delete('', 'v1', 'serviceaccounts', annotations={'foo': 'bar'})
async def foo(**_):
    pass
```

Then I apply:
```
apiVersion: v1
kind: ServiceAccount
metadata:
  annotations:
    foo: bar
  name: test
  namespace: default
```

and the finalizer is applied.

I then run:
`kubectl patch sa test -p '{"metadata": {"annotations": {"foo": "baz"}}}'`

and the finalizer is removed.

psycho-ir:
> Hey!
> 
> I also can't reproduce.
> 
> I have the operator as:
> 
> ```
> import kopf
> 
> @kopf.on.delete('', 'v1', 'serviceaccounts', annotations={'foo': 'bar'})
> async def foo(**_):
>     pass
> ```
> 
> Then I apply:
> 
> ```
> apiVersion: v1
> kind: ServiceAccount
> metadata:
>   annotations:
>     foo: bar
>   name: test
>   namespace: default
> ```
> 
> and the finalizer is applied.
> 
> I then run:
> `kubectl patch sa test -p '{"metadata": {"annotations": {"foo": "baz"}}}'`
> 
> and the finalizer is removed.

Right 😕,

I probably did a mistake in my tests, all the scenarios are working a charm. 
Sorry to bother.

nolar:
PS: The solution in general is fine. 

I suggest that we merge it now, and release as 0.21rcX (x=3..4 [or so](https://github.com/zalando-incubator/kopf/releases)), together with lots of other bugfixes/refactorings/improvements, and test them altogether.

psycho-ir:
> PS: The solution in general is fine.
> 
> I suggest that we merge it now, and release as 0.21rcX (x=3..4 [or so](https://github.com/zalando-incubator/kopf/releases)), together with lots of other bugfixes/refactorings/improvements, and test them altogether.

👍

nolar:
Pre-released as [kopf==0.21rc3](https://github.com/zalando-incubator/kopf/releases/tag/0.21rc3) — but beware of other massive changes in rc1+rc2+rc3 combined (see [Releases](https://github.com/zalando-incubator/kopf/releases)).

PR #176, opened 2019-08-08
Switch to async i/o for internal API communication

### LINKED ISSUE #142: Slow or hanging process termination

nolar:
> Issue : fixes #142, replaces #152

## Problem

There is no easy way to terminate the blocking `read()` operation on a socket. One way is a dirty hack with the OS-level process signals, which interrupt the I/O operations on low level (OS&libc&co) — see #152.

## Solution

The proper solution, however, is to use async i/o inside of the async app.

**This PR** converts all i/o to/from Kubernetes API to `aiohttp`. It is already present in the dependencies indirectly.

This efficiently removes pykube-ng (or any other clients) from the Kopf's core. They are not much needed, as the main purpose of the client libraries is to provide a convenient DSL (domain-specific language) for the Kubernetes objects manipulation. In Kopf, all manipulation is unified, and which objects are being handled is not so important.

## Implementation

**For authentication,** both pykube-ng and the Kubernetes client remain as the dependencies: for config parsing, token retrieval, including the edge cases of GCE tokens, local command executors, so on. Kopf piggybacks on pykube-ng/client for this purpose.

**For testing,** [aresponses](https://github.com/CircleUp/aresponses) is used instead of mocked `requests`. It runs an actual web server locally, and intercepts all aiohttp outcoming requests to be redirected to that web-server.

This switch led to almost full rewrite of all tests for `kopf.clients` module (all API communication) — which makes a half of this PR's size (while keeping the same semantics of the tests).

## Types of Changes

- Refactor/improvements

## Remaining TODOs

* Authentication:
  * [ ] Check piggybacking on Pykube with tokens (now: only checked with certs@minikube).
  * [ ] Check piggybacking on Pykube with username/password.
  * [ ] Check if it works with GCE (reported as broken in Kopf+Pykube).
  * [ ] Implement piggybacking on the Kubernetes Client when pykube is absent.
  * [ ] Require that at least one of the is installed, make pykube optional.
  * [ ] Add automatic re-login, mostly for refreshment of GCE tokens?
* Embeddable operators:
  * [ ] Accept KubeConfig/Confguration objects from the user, instead of `kopf.login()`.
  * [ ] Document login procedured for embeddable operators.
* [ ] Improve the exception reporting from within aresponses's handlers (now: it is shown in the teardown section).
* [ ] Add version of self to User-Agent.
* [ ] Ensure that cluster-scoped objects also do work (different namespace handling).

## Review

- [ ] Tests
- [ ] Documentation


brutus333:
Hello and thank you for developing Kopf! 

I implemented a daemonset like object using Kopf and I faced some difficulties. But I don't want to talk about difficulties here (I'll open a issue at some point probably), I only want to ask you something that it is not very clear for me and it is related to this PR:

I have a rather weak understanding of async in Python so please bear with me, the question is:
if we have to call only async libraries in an async app and if a kopf handler is an async app (if I got this right) then calling sync kubernetes client libraries in a kopf handler is the right thing to do? 

Shouldn't we need to use an async kubernetes client library?

Thank you in advance,
Virgil

brutus333:
While digging trough the code I found the reference to https://pymotw.com/3/asyncio/executors.html so this is answering my question.

nolar:
Closed in favour of #226 (auth) + #227 (aiohttp).

Issue #184, opened 2019-08-21
Embedded Example Broken


cliffburdick:
## Expected Behavior
Example working

## Actual Behavior
Wrong syntax

## Steps to Reproduce the Problem

1. Run the embedded example here:
https://kopf.readthedocs.io/en/latest/embedding/

2. You will see ```TypeError: __init__() got an unexpected keyword argument 'target_fn'``` since the actual parameter is "target".

3. Once that's fixed, if you make ```def main``` into ```if __name__ == '__main__'```, you will see this:

```
/usr/lib/python3/dist-packages/requests/__init__.py:80: RequestsDependencyWarning: urllib3 (1.25.3) or chardet (3.0.4) doesn't match a supported version!
  RequestsDependencyWarning)
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.7/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "test.py", line 11, in kopf_thread
    loop = asyncio.get_event_loop()
  File "/usr/lib/python3.7/asyncio/events.py", line 644, in get_event_loop
    % threading.current_thread().name)
RuntimeError: There is no current event loop in thread 'Thread-1'.
```

4. After trying to hack around that by adding ```asyncio.set_event_loop(asyncio.new_event_loop())``` before the loop assignment, I get:
```AttributeError: module 'kopf' has no attribute 'operator'```

## Specifications

  - Platform: Ubuntu 18.04
  - Kubernetes version: None
  - Python version: 3.7.4

Also, is it possible to not require python 3.7?

nolar:
@cliffburdick Regarding Python>=3.7 requirement: Kopf will not function with 3.6 anyway, as it uses some 3.7+ features and modules actively in its core (e.g. `contextvars`, but there are others). See #74 on reasoning.

nolar:
@cliffburdick The `target=` is fixed now in the docs. And a real runnable example is added. So, I assume, this issue is done. If not, feel free to re-open.

PR #185, opened 2019-09-03
Add an example for embedded operator (and stop/ready flags)

nolar:
Add stop/ready flags to the operator coroutines, and add an example for the embedded operator, which uses these flags.

> Issue : #184, follow-up for #156.

## Description

The stop-flag is used to signal the operator (from the outside caller) that it should terminate gracefully.

The ready-flag is used to signal the outside caller (from the operator) that the operator has initialised and is ready to work.

This bi-directional communication slightly simplifies the orchestration of the operators when they are embedded into arbitrary applications (i.e. not called via `kopf run`). This could be also done by cancelling the task of the operator, but that would require few extra lines to operate it (checking for the cross-thread flag and finishing the asyncio task), which are already present inside of the operator coroutines — so why not utilise that?

An example is added to demonstrate how an embedded operator can look like. 

Also, the docs were fixed for the embedded operator.

## Types of Changes

- Documentation / non-code

### DIFF
--- a/kopf/reactor/running.py
+++ b/kopf/reactor/running.py
@@ -289,8 +289,8 @@ async def _stop_flag_checker(
 
     # Wait until one of the stoppers is set/raised.
     try:
-        aws = [signal_flag] + ([] if stop_flag is None else [_wait_flag(stop_flag)])
-        done, pending = await asyncio.wait(aws, return_when=asyncio.FIRST_COMPLETED)
+        flags = [signal_flag] + ([] if stop_flag is None else [_wait_flag(stop_flag)])
+        done, pending = await asyncio.wait(flags, return_when=asyncio.FIRST_COMPLETED)
         future = done.pop()
         result = await future
     except asyncio.CancelledError:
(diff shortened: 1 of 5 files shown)


kopf/reactor/running.py
```diff
@@ -265,9 +278,21 @@ def run(
...
+        stop_flag: Optional[Flag],
+):
+    # TODO: collect the readiness of all root tasks instead, and set this one only when fully ready.
+    # Notify the caller that we are ready to be executed.
+    await _raise_flag(ready_flag)
+
+    # Wait until one of the stoppers is set/raised.
     try:
-        result = await should_stop
+        aws = [signal_flag] + ([] if stop_flag is None else [_wait_flag(stop_flag)])
```
samurang87: aws?

nolar: Awaitables, as in the docs: https://docs.python.org/3/library/asyncio-task.html#asyncio.wait

samurang87: "awaitables" looks better than "aws" IMO :P

nolar: @samurang87 Renamed it to `flags`.

Issue #194, opened 2019-10-01
Add type-hints all over the code


nolar:
## Motivation

For the convenience of the external contributors, especially the new starters, it would be nice to have type hints for all function/classes signatures, so that at every place in the codebase it is clear what is what, which methods it has, and what can be done with it.

This is especially important since there is a terminology conflict both within the framework (derived from the K8s conflict: `kind: Event` k8s-events vs. `GET .../?watch=true` watch-events), and between the framework and K8s (e.g. "bodies" as the body of the requests, sometimes called "objects"). And all of this is mentioned both directly, and in generic classes (e.g. a queue of k8s-events). 

Beside of that, type-hinting requires extra imports in the modules (for type references), which, in turn, requires proper design of the abstractions and layers. Currently, there are few implicit cyclic cross-dependencies. The type-hinting will make them explicit (though requires moving then across the codebase).

## Benefits:

* The type-hinting can clarify what is used in which context. 
* The IDEs can benefit from it providing the proper autocompletion. 
* Extended type-hinting (such as TypedDict) can also expose the actual data structures used, despite them being just of `dict` type (e.g. bodies' metadata fields and their types).
* Extra strictness on types can prevent bugs, especially in the edge cases (such as None vs. str for some values).
* Better design in general, less corrupted cross-dependencies.
* Utilise new Python 3.7+ features at full strength (enums, etc).

## Steps:

* [x] Extract `Resource` to `kopf.structs` as a separate reusable class.
* [x] Extract `Registries` to `kopf.structs` (they are structs after all, not a reactor part).
* [x] Create `kopf.toolkit` for unrelated helpers and tools (incl. `kopf.testing` and hierarchy management).
* [x] Add lazy resolvers for the body dict for spec, metadata, status.
* [x] Declare the callback signatures. for handlers & lifecycles.
* [x] Hide the pykube-ng's/requests's Response, return items+resource_version in `list_objs()`.
* [x] Declare the API body parts and their relations.
* [x] Add type-hints to all function/class signatures.
* [x] Enable `mypy --strict` mode in CI.
* [ ] Expose all user-facing handler-related type declarations from `kopf.__init__`.

## Related:

* Linters #47.

PR #197, opened 2019-10-01
Remove an abstraction leak of a Response object — for type-hinting

nolar:
Refactor clients-to-reactor communication to prepare for the overall type-hinting.

> Issue : #194 

## Description

Since the switch to pykube-ng (#71 #110), there was an [_abstraction leak_](https://en.wikipedia.org/wiki/Leaky_abstraction) in the list-fetching client wrapper: the `requests.Response` object was returned, thus exposing the internal implementation of a sibling module.

In this PR, the leak is removed, and the function now exposes only the data needed. It also takes the hacky data recovery (missing `apiVersion` & `kind` in the lists) from the Kopf's reactor into the client wrapper.

Having this in place will also simplify future switch to other HTTP clients, such as asyncio (see #176), as there is no such a specific response object in other implementations.

The function is intentionally renamed: to break things if someone for some reason uses this _private_ function in their code — since its signature is now different (another result type).

## Types of Changes

- Refactor/improvements

**There are no behavioural changes. Only the code refactoring.** The tests show that all the protocols and interfaces are kept in place — except for the imports of the internal modules (not exposed via `kopf/__init__.py`).

### DIFF
--- a/kopf/clients/fetching.py
+++ b/kopf/clients/fetching.py
@@ -41,7 +41,7 @@ def read_obj(*, resource, namespace=None, name=None, default=_UNSET_):
         raise
 
 
-def list_objs(*, resource, namespace=None):
+def list_objs_rv(*, resource, namespace=None):
     """
     List the objects of specific resource type.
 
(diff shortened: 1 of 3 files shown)


kopf/clients/fetching.py
```diff
@@ -58,4 +58,16 @@ def list_objs(*, resource, namespace=None):
     cls = classes._make_cls(resource=resource)
     namespace = namespace if issubclass(cls, pykube.objects.NamespacedAPIObject) else None
     lst = cls.objects(api, namespace=pykube.all if namespace is None else namespace)
-    return lst.response
+    rsp = lst.response
+
+    items = []
+    resource_version = rsp.get('metadata', {}).get('resourceVersion', None)
+    for item in rsp['items']:
+        # FIXME: fix in pykube to inject the missing item's fields from the list's metainfo.
```
jonathanbeber: I didn't find an issue on `pykube` for that, hasn't it been reported yet?

nolar: Nope. Not sure it should be there either. It is a way how K8s API works.

PR #217, opened 2019-10-28
Convert client wrappers to coroutines, use thread pools for actual calls

nolar:
> Issue : #142

## Description

The _client wrappers_ are used to isolate the framework's reactor from the specific HTTPS API calls needed to talk to Kubernetes, and from the ways of doing so. They provide a limited set of operations needed only, and no more than that (no general-purpose DSL for resource manipulation).

In this PR, all remaining synchronous client wrappers are converted to async coroutines, while they remain Pykube-based internally (`pykube-ng` is synchronous, as `requests` are synchronous).

This is needed both for the switch to aiohttp (e.g. #176), and for the re-authentication activities (few issues).

The synchronous calls are moved few levels down the stack, and put into the thread-pool executors. 

Previously, these calls were sync inside of the async event loop, so they were blocking. It was not a big problem, as most of them are executed only once or twice per run in the very beginning; the frequent resource-watching & resource-patching & event-posting calls were already executed in the thread pools. But it now becomes a problem since all API calling methods cannot be so diverse anymore, and have to be uniform (specifically, async).

## Types of Changes

- Refactor/improvements

### DIFF
--- a/kopf/clients/classes.py
+++ b/kopf/clients/classes.py
@@ -1,17 +1,24 @@
+import asyncio
+import functools
 from typing import Type
 
 import pykube
 
+from kopf import config
 from kopf.clients import auth
 from kopf.structs import resources
 
 
-def _make_cls(
+async def _make_cls(
         resource: resources.Resource,
 ) -> Type[pykube.objects.APIObject]:
 
+    loop = asyncio.get_running_loop()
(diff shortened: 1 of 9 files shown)


nolar:
Zincr-bot, ping.

Issue #219, opened 2019-11-05
kopf fails to run on Windows


damc-dev:
## Expected Behavior
kopf should run example

## Actual Behavior
Fails with NotImplementedError from asyncio
```
[2019-11-05 16:05:13,664] kopf.clients.auth    [DEBUG   ] Pykube is configured via kubeconfig file.
[2019-11-05 16:05:13,737] kopf.clients.auth    [DEBUG   ] Client is configured via kubeconfig file.
[2019-11-05 16:05:13,808] kopf.engines.peering [WARNING ] Default peering object not found, falling back to the standalone mode.
Traceback (most recent call last):
  File "c:\python37\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "c:\python37\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Python37\Scripts\kopf.exe\__main__.py", line 7, in <module>
  File "c:\python37\lib\site-packages\click\core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "c:\python37\lib\site-packages\click\core.py", line 717, in main
    rv = self.invoke(ctx)
  File "c:\python37\lib\site-packages\click\core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "c:\python37\lib\site-packages\click\core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "c:\python37\lib\site-packages\click\core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "c:\python37\lib\site-packages\kopf\cli.py", line 30, in wrapper
    return fn(*args, **kwargs)
  File "c:\python37\lib\site-packages\kopf\cli.py", line 63, in run
    peering_name=peering_name,
  File "c:\python37\lib\site-packages\kopf\reactor\running.py", line 44, in run
    peering_name=peering_name,
  File "c:\python37\lib\asyncio\base_events.py", line 579, in run_until_complete
    return future.result()
  File "c:\python37\lib\site-packages\kopf\reactor\running.py", line 77, in operator
    ready_flag=ready_flag,
  File "c:\python37\lib\site-packages\kopf\reactor\running.py", line 157, in spawn_tasks
    loop.add_signal_handler(signal.SIGINT, signal_flag.set_result, signal.SIGINT)
  File "c:\python37\lib\asyncio\events.py", line 540, in add_signal_handler
[...]

PR #220, opened 2019-11-06
Fix Windows Compatibility

damc-dev:
When running kopf on Windows it fails due to NotImplementedError thrown by add_signal_handler method in asyncio

> Issue : #219 

# Description
In the asyncio docs it appears that the [add_signal_handler](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.add_signal_handler) method is only available in Unix which is why it throws a NotImplementedError when kopf is run on Windows.
This PR should fix Windows compatibility by catching NotImplementedError thrown by asyncio and continuing.

# Types of Changes
* Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/kopf/reactor/running.py
+++ b/kopf/reactor/running.py
@@ -163,8 +163,13 @@ async def spawn_tasks(
 
     # On Ctrl+C or pod termination, cancel all tasks gracefully.
     if threading.current_thread() is threading.main_thread():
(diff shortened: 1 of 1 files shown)


nolar:
Released as [`kopf==0.23rc1`](https://github.com/zalando-incubator/kopf/releases/tag/0.23rc1).

@damc-dev Thank you for you contribution! 😉

Issue #187, opened 2019-09-16
Running the Tutorial against an EKS cluster


jicowan:
## Expected Behavior
I would like to run the kopf tutorial on an EKS cluster instead of minikube.  

## Actual Behavior
When I run `kopf run <file>.py --verbose` I get an unauthorized error: 
`pykube.exceptions.HTTPError: customresourcedefinitions.apiextensions.k8s.io "kopfpeerings.zalando.org" is forbidden: User "system:anonymous" cannot get resource "customresourcedefinitions" in API group "apiextensions.k8s.io" at the cluster scope`

I see that kopf is using the pykube library which doesn't natively support IAM authentication.  Is there a way to make this work without using a bearer token for authentication?

## Specifications

  - Platform: EKS 
  - Kubernetes version: *(use `kubectl version`)* 1.13
  - Python version: *(use `python --version`)* 3.7.4
  - Python packages installed: *(use `pip freeze --all`)* 
aiohttp==3.6.0
aiojobs==0.2.2
async-timeout==3.0.1
attrs==19.1.0
cachetools==3.1.1
certifi==2019.9.11
chardet==3.0.4
Click==7.0
google-auth==1.6.3
idna==2.8
iso8601==0.1.12
kopf==0.21
kubernetes==10.0.1
multidict==4.5.2
oauthlib==3.1.0
pip==19.0.3
pyasn1==0.4.7
pyasn1-modules==0.2.6
pykube-ng==19.9.2
python-dateutil==2.8.0
PyYAML==5.1.2
requests==2.22.0
requests-oauthlib==1.2.0
rsa==4.0
setuptools==40.8.0
six==1.12.0
urllib3==1.25.3
websocket-client==0.56.0
yarl==1.3.0

```
```

nolar:
Kopf uses `pykube-ng` (which is a fork of `pykube`). Beside token-bearer, it also supports client certificate, username-password, and some other auth methods. At least, it is [in the 
code](https://github.com/hjacobs/pykube/blob/19.9.2/pykube/http.py#L128-L134).

I will try to reproduce the error locally for EKS. Maybe @hjacobs can help here too.

---

[...]

PR #221, opened 2019-11-06
Refactoring 4/3: In-memory handler outcomes & persistent state merges

nolar:
Code refactoring (part 4 of 3): split reactor state to handlers' outcomes and actual state persistence.

> Issue : preparation for #59 #187 #142
> Preceded: #209 (part 1/3), #210 (part 2/3), #211 (part 3/3).

## Description

Having terminology fixed in #209 and the "state" word released for the reactor's state persistence, we now improve the state management itself.

**The purpose** is to separate the state persistence in the object's status stenza from the handler's outcomes. Specifically, to make it possible to keep the state of the handlers that are not resource-related — e.g. activity handlers, which are handled completely in-memory.

For this, a new entity is introduced: an *outcome*, which is conceptually close to asyncio/threading futures (result+exception bundled), but have handler-specific aspects: delays, temporary or permanent errors, etc.

The state persistence is then refactored to include the math of merging the existent state with the new outcomes, and persisting the new resulting state back to the objects.

The resource-specific state persistence is then moved upward in the call stack: from the generic handlers, to the resource-specific handling. The generic handlers now return the in-memory *outcomes* only, and persist nothing.

_See the list of commit with step-by-step explanation of changes._

## Types of Changes

- Refactor/improvements


ancelotpinto:
lgtm

PR #222, opened 2019-11-08
Configure reactions to errors in the handlers, per-handler

### LINKED ISSUE #16: Do not show the stacktraces on the retry/fatal exceptions
The retry-/fatal-exceptions are the signals for the framework to stop handling or to retry the handler. They are not the regular "unexpected" errors, since they are expected.

As such, the stacktraces should not be printed. Yet the messages of the exceptions should be printed/logged.

For other regular exceptions, which are by design "unexpected", the stacktraces should be printed/logged as usually (similar as if they happen inside of a thread/greenlet, and the main process continues).

nolar:
Extra keywords for the handlers to override the default exception-reacting behaviour (timeouts, retries, cool-down periods, arbitrary error interpretations).

> Issue : closes #16

## Description

With this PR, the error handling mode can be configured by handler (for arbitrary exceptions):

* Treat them as permanent: ``errors=kopf.ErrorsMode.PERMANENT``
* Treat them as temporary for N attempts: ``retries=3``
* Treat them as temporary for some time: ``timeout=60`` (exists already)
* Ignore them -- same as permanent, with better logging: ``errors=kopf.ErrorsMode.IGNORED``

The current and historic behaviour (the new default) is to treat all errors as temporary until the timeout is reached, or forever until the resource is deleted (if timeout is not set).

Also, a _cooldown_ period between retries can be configured. Previously, it was hard-coded to 60 seconds.

The example with exceptions is changed accordingly to demo the now extended usage.

*See the list of commits for detailed step-by-step changes.*

## Types of Changes

- Refactor/improvements

## Review
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/docs/errors.rst
+++ b/docs/errors.rst
@@ -65,11 +65,11 @@ Regular errors
 
 Kopf assumes that any arbitrary errors
 (i.e. not `TemporaryError` and not `PermanentError`)
-are the environment issues and can self-resolve after some time.
+are environment issues and can self-resolve after some time.
 
 As such, as a default behaviour,
 Kopf retries the handlers with arbitrary errors
-infinitely until they either succeed or fail permanently.
+infinitely until the handlers either succeed or fail permanently.
 
 The reaction to the arbitrary errors can be configured::
 
--- a/docs/errors.rst
+++ b/docs/errors.rst
@@ -60,15 +60,30 @@ is no need to retry over time, as it will not become better::
             raise kopf.PermanentError("The object is not valid anymore.")
 
 
-
 Regular errors
 ==============
 
-Any other exceptions behave as either the retriable (default) or fatal,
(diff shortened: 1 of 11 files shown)


docs/errors.rst
```diff
@@ -60,15 +60,30 @@ is no need to retry over time, as it will not become better::
...
 
-
 Regular errors
 ==============
 
-Any other exceptions behave as either the retriable (default) or fatal,
-depending on the settings.
+Kopf assumes that any arbitrary errors
+(i.e. not `TemporaryError` and not `PermanentError`)
+are the environment issues and can self-resolve after some time.
```
dneuhaeuser-zalando: should be: "are environment issues" without "the"

nolar: Fixed.

PR #228, opened 2019-11-13
Liveliness/readiness probes

### LINKED ISSUE #18: Add health checks
## Expected Behavior

The operator is restarted by Kubernetes if it becomes irresponsive.

## Actual Behavior

The operator can stuck for any reason (e.g. bugs), and nobody will notice — except as by no reaction to the added/deleted objects.

  1. Put a synchronous `time.sleep(300)` anywhere in the async handler (`async def ...`).
  1. Let it run.
  1. Observe how the operator is blocked for 5 mins for all objects.

nolar:
Add a rudimentary liveness probe via an HTTP endpoint

> Issue: closes #18

## Description

Having Kopf's core now fully in `asyncio` and `aiohttp` (#227), we can add a little web server to serve the liveness probes of the operator. 

It can be used by Kubernetes to see if the process is at all alive.

By default, there is no useful information reported, and only HTTP status codes can be used to track the liveness — e.g. 200 OK.

Optionally, the reported information can be collected from the `@kopf.on.probe` handlers, e.g. containing the app-specific metrics (see docs & examples). Some basic DoS/DDoS protection is made. But the endpoint is supposed to be internal, so it is only a precaution for accidental misconfiguration.

## Types of Changes

## Review
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/docs/deployment.rst
+++ b/docs/deployment.rst
@@ -58,6 +58,10 @@ For this, create the deployment file:
 Please note that there is only one replica. Keep it so. If there will be
 two or more operators running in the cluster for the same objects,
 they will collide with each other and the consequences are unpredictable.
+In case of pod restarts, only one pod should be running at a time too:
+use ``.spec.strategy.type=Recreate`` (see the documentation_).
+
+.. _documentation: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#recreate-deployment
 
 Deploy it to the cluster:
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -31,6 +31,7 @@ Kopf: Kubernetes Operators Framework
    hierarchies
    async
    loading
+   probing
    peering
    scopes
    errors
--- a/docs/probing.rst
+++ b/docs/probing.rst
@@ -1,0 +1,51 @@
+=============
+Health-checks
+=============
+
+Kopf provides a minimalistic HTTP server to report its health status.
+
+By default, no endpoint is configured, and no health is reported.
+To specify an endpoint to listen for probes, use :option:`--liveness`:
+
+.. code-block:: bash
+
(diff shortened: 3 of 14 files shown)


dneuhaeuser-zalando:
Liveness probes are somewhat problematic. This is covered both in our internal documentation as well as in this blog post Henning made: https://srcco.de/posts/kubernetes-liveness-probes-are-dangerous.html

As an operator is probably not likely to get overloaded in the same way an API is, I think this may be not as big a concern. I also think that a liveness probe is a legitimate solution to mitigate the blocked event loop case. However there is still some risk here. Therefore I think a warning about this in the docs would be appropriate. It would also be good, to have documentation on how to monitor the operator to identify the blocked event loop case.

nolar:
@dneuhaeuser-zalando Added few notes.

It is true that the API approach is not applicable for the operators. Currently, only one operator pod must be running at a time. Two or more pods will conflict for the resources in strange ways. 

I had few ideas how to run the operators with multiple pods by "assigning" (acquiring, locking) the resources to the individual operators for short time. But that is not actually a problem at the moment. I don't know of any operators of such a scale that load distribution would make sense at all.

And for this reason, there can be no readiness probe, only the liveness probe. And even the liveness probe is minimalistic — just enough to notice that the operator is dead and not responding at all, so should be restarted. 

There is currently no self-checks on individual aspects of the operator. It can be so that the web server is alive, but one of the watch-streams is stuck on the API calls. This will not be detected. Maybe, later, I can add some advanced self-checks (as soon as it becomes clear what is worth checking in case of operators — based on our team's experience of using the rudimentary liveness checks).

The probing handlers and the returned JSON structs are just a fancy side-effect. Can be used for the domain metrics monitoring — I will try that later.

Issue #112, opened 2019-06-14
Is resume called before create/update?


Jc2k:
Hi!

I have a few questions about resume handlers:

 * Are there any guarantees about call order with `.on.resume` vs update and create?
 * Will resume get called always, or could sometimes update be called instead?
 * Is it possible to know that kopf has finished processing the current set of resumes? As in, has it finished start up and is now watching and (crucially) waiting for events?
 * Is there any way where I could see a resume event after i've processed a delete event?

I'd like to populate a cache with all instances of my CRD at start up, then keep it up to date by following the created/update/delete events. I'd like to be able to know that my cache is in a good state before i start processing events so that I can reference it. Is this doable?

nolar:
Neither before, nor after, but _as part_ of create/update.  Usually, the handlers are called in the order they are declared. Resume handlers are then just _mixed-in_ into the list of handlers to call, if this is an operator restart.

The order of handlers can be additionally controlled with the _handlers lifecycles_. It seems, they are not documented (accidentally); but they ARE the public interface. These ones: https://kopf.readthedocs.io/en/latest/packages/kopf.reactor.lifecycles/. Usage example:

```python
import kopf

kopf.set_default_lifecycle(kopf.lifecycles.one_by_one)

@kopf.on.whatever...
```

Or you can make your own callback and control the order as you wish (and store the state in the status).

I will document the handler ordering control a bit later.

nolar:
Both resume+update, or resume+create handlers will be called. However, if they point to the same function, they will be de-duplicated, and that function will be called only once. E.g.:

```python
import kopf

@kopf.on.resume('zalando.org', 'v1', 'kopfexamples')
@kopf.on.create('zalando.org', 'v1', 'kopfexamples')
@kopf.on.update('zalando.org', 'v1', 'kopfexamples')
def fn(spec, **kwargs):
    print(spec)
```

The `fn` function will be called only once per event, be that creation or update or operator restart (i.e. object resuming).

```python
import kopf

@kopf.on.resume('zalando.org', 'v1', 'kopfexamples')
def fn1(spec, **kwargs):
    print(spec)

@kopf.on.create('zalando.org', 'v1', 'kopfexamples')
def fn2(spec, **kwargs):
    print(spec)

@kopf.on.update('zalando.org', 'v1', 'kopfexamples')
def fn3(spec, **kwargs):
    print(spec)
```

[...]

PR #234, opened 2019-11-14
Use `memo` for arbitrary per-resource payload during operator lifetime

nolar:
> Issue : #112 

## Description

Since the per-resource in-memory runtime-only containers are already present in the core for internal purposes, it is easy to expose `memo` kwarg to the handlers, so that the can store arbitrary information or objects about resources.

Same as for the internal use, this should be the objects that are scoped to the operator life time, fully lost on restarts, i.e. not persistent. E.g. threads, tasks, locks, events. Maybe some regular scalars, if their meaning is also scoped to the operator process.

Example:

```python
import kopf

@kopf.on.create('zalando.org', 'v1', 'kopfexamples', cooldown=10)
def create_fn_1(memo, **kwargs):
    print(f'{memo.xyz}')

def create_fn_2(memo, **kwargs):
    memo.xyz = 100
```

In this example, `create_fn_1` raises AttributeError on the 1st attempt, then goes to `create_fn_2`, and then retries `create_fn_1` in 10 seconds with the `.xyz` field already set.

Both dictionary and object interfaces are supported.

_The documentation is intentionally limited — let's first see how it will be used, and what can be the examples._

## Types of Changes

## Review
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/kopf/structs/containers.py
+++ b/kopf/structs/containers.py
@@ -8,24 +8,24 @@ It is used internally to track allocated system resources for each Kubernetes
 object, even if that object does not show up in the event streams for long time.
 """
 import dataclasses
-from typing import MutableMapping
+from typing import MutableMapping, Dict, Any
 
 from kopf.structs import bodies
 
 
-class ObjectDict(dict):
+class ObjectDict(Dict[Any, Any]):
     """ A container to hold arbitrary keys-fields assigned by the users. """
 
-    def __setattr__(self, key, value):
+    def __setattr__(self, key: str, value: Any) -> None:
         self[key] = value
 
-    def __delitem__(self, key):
+    def __delitem__(self, key: str) -> None:
         try:
             del self[key]
         except KeyError as e:
             raise AttributeError(str(e))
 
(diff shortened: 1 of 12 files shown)


kopf/structs/containers.py
```diff
@@ -13,9 +13,33 @@
...
+    """ A container to hold arbitrary keys-fields assigned by the users. """
+
+    def __setattr__(self, key, value):
+        self[key] = value
+
+    def __delitem__(self, key):
+        try:
+            del self[key]
+        except KeyError as e:
+            raise AttributeError(str(e))
```
dneuhaeuser-zalando: Doesn't `del self[key]` call `self.__delitem__(key)`? Could it be that you intended to implement [`__delattr__`](https://docs.python.org/3/reference/datamodel.html#object.__delattr__)? If not, I do think that implementing the latter would be a good idea for consistency reasons.

nolar: Oops. That was `__delattr__`, of course. Fixed. And few more tests added for deletion of the attrs & keys.

Issue #242, opened 2019-11-19
ValueError: Cannot combine AUTHORIZATION header with AUTH argument or credentials encoded in URL


chungktran:
## Expected Behavior
kopf to run normall

## Actual Behavior
kopf throws error `ValueError: Cannot combine AUTHORIZATION header with AUTH argument or credentials encoded in URL`.

Please note that this error occurs when using kopf with k3s. However, I don't think k3s is the problem here but rather that it's using username/password authentication instead token.

## Steps to Reproduce the Problem

  1. Run kopf with or without `--dev` flag

## Specifications

  - Platform: Linux
  - Kubernetes version: v1.16.0
  - Python version: 3.7.4
  - Python packages installed: *(use `pip freeze --all`)*

```
Package           Version
----------------- ---------
aiohttp           3.6.2
aiojobs           0.2.2
async-timeout     3.0.1
attrs             19.3.0
cachetools        3.1.1
certifi           2019.9.11
chardet           3.0.4
Click             7.0
google-auth       1.7.1
idna              2.8
iso8601           0.1.12
kopf              0.23rc4
kubernetes        10.0.1
multidict         4.5.2
oauthlib          3.1.0
pip               19.3.1
pyasn1            0.4.7
pyasn1-modules    0.2.7
pykube-ng         19.10.0
python-dateutil   2.8.1
PyYAML            5.1.2
requests          2.22.0
requests-oauthlib 1.3.0
rsa               4.0
setuptools        41.6.0
six               1.13.0
typing-extensions 3.7.4.1
urllib3           1.25.7
websocket-client  0.56.0
wheel             0.33.6
yarl              1.3.0
```

Stack Trace
```
$ kopf run release.py
[2019-11-18 11:35:20,726] kopf.reactor.activit [INFO    ] Initial authentication has been initiated.
[2019-11-18 11:35:20,779] kopf.activities.auth [INFO    ] Handler 'login_via_pykube' succeeded.
[2019-11-18 11:35:20,806] kopf.activities.auth [INFO    ] Handler 'login_via_client' succeeded.
[2019-11-18 11:35:20,807] kopf.reactor.activit [INFO    ] Initial authentication has finished.
Traceback (most recent call last):
  File "/home/ctran/.venv/release/bin/kopf", line 8, in <module>
    sys.exit(main())
  File "/home/ctran/.venv/release/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
[...]

PR #243, opened 2019-11-19
Fix an issue with mistakenly added Bearer auth in addition to Basic auth

nolar:
> Issue : #242

## Description

`Authorization: Bearer` header was sent (without a token!) because there was a schema defined by default (`"Bearer"`), which should not be there (`None`).

This caused problems when Basic auth (username+password) was used — it could not co-exist with `Authorization: Bearer` header.

## Types of Changes


dneuhaeuser-zalando:
Approved because I understand it's somewhat urgent but ideally this should have a test, to ensure this fixes the problem and to prevent a regression.

nolar:
The tests for this case are added post-factum in #244 (with other auth/ssl tests).

PR #244, opened 2019-11-19
Support raw data for  SSL certs, pkeys, and CA

nolar:
> Issue : follow-up for #226, extra fix for #242 

## Description

In #226, custom authentication was added, with CA/SSL-cert/SSL-pkey accepted both as path and data — similar to how it is stored in the `~/.kube/config`.

However, data fields were ignored and not handled in any way.

This PR should fix this.

## Types of Changes

### DIFF
--- a/kopf/clients/auth.py
+++ b/kopf/clients/auth.py
@@ -1,10 +1,11 @@
 import base64
 import functools
+import os
 import ssl
 import tempfile
 import warnings
 from contextvars import ContextVar
-from typing import Optional, Callable, Any, TypeVar, Dict, cast
+from typing import Optional, Callable, Any, TypeVar, Dict, Iterator, Mapping, cast
 
 import aiohttp
 
@@ -85,6 +86,7 @@ class APISession(aiohttp.ClientSession):
     """
     server: str
     default_namespace: Optional[str] = None
+    _tempfiles: "_TempFiles"
 
     @classmethod
     def from_connection_info(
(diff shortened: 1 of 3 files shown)


dneuhaeuser-zalando:
As I already wrote in the other PR tests would be good. Also I'm suspicious of the temporary file handling. You're not cleaning them up, so I guess they leak and you eventually run out of memory/disk space, right?

nolar:
@dneuhaeuser-zalando  Indeed, they will be leaking. I've taken that from the [pykube-ng's code](https://github.com/hjacobs/pykube/blob/19.10.0/pykube/config.py#L287-L289). Kubernetes client does the same ([1](https://github.com/kubernetes-client/python-base/blob/a2d1024524de78b62e5f9aa72d34cb9ea9de2b97/config/kube_config.py#L436-L444), [2](https://github.com/kubernetes-client/python-base/blob/a2d1024524de78b62e5f9aa72d34cb9ea9de2b97/config/kube_config.py#L69-L72)), just [cleans the files at exit](https://github.com/kubernetes-client/python-base/blob/a2d1024524de78b62e5f9aa72d34cb9ea9de2b97/config/kube_config.py#L62-L63). SO, the approach is as secure, as anything else in this field.

I have now fixed it to also purge it — just not `atexit`, but on the garbage collection of an API session, which probably prevents the leakages for long-running operators. I.e., the files are kept as long as they are actually used, not as long as the operator runs.

And there are now few tests. But they do not cover everything — I didn't find how to extract a parsed client certificate or even its existence from an SSLContext object. It is more or less testable with CAs.

Issue #98, opened 2019-06-02
Filter by arbitrary callback function


nolar:
In some cases, it might be needed that only a subset of resources is served and handled. Some of such filtering are suggested in #45 #58, but it can be anything arbitrary.

Without such filtering, the logs will be cluttered by the resource events/causes that are of no interest to the operator.

For this, a callback function can be specified on the handler declaration:

```python
KNOWN_PODS = {}

@kopf.on.event('', 'v1', 'pods'
               when=lambda event, uid, **_:  event=='ADDED' and uid not in KNOWN_PODS)
def new_pod_created(**kwargs):
    pass
```

The callback should be called via the same invocation protocol as the handlers (see `kopf.reactor.invocation`), i.e. with all the handler's kwargs — be that an event-handler or a cause-handler (different sets of kwargs).

If the callback returns true, then the object is handled as normally. When it returns false, the handler is ignored, maybe the whole event is ignored (if there are no other handlers) — and therefore **produces no single line of logs at all**.

No additional kwargs are injected for such filtering.

----

Related: 

* #30 for silent handlers (no status/progress storage).
* #84 for pods support (but can be tested with another custom resource).
* #58 for filtering by owner references (aka parents).
* #45 for filtering by labels/annotations.

Checklist:

* [x] Acceptance criteria:
  * [x] Event handler registered with the filtering callback function/lambda/partial.
  * [x] Callback is invoked every time an event is arrived and/or cause is detected.
  * [x] Events/handlers are filtered by the result of this callback invocation.
  * [x] No logs about handler or event/cause are produced is the handler was filtered out.
  * [x] Minimum overhead and not extra calls if the callback is not specified (the default).
* [x] Tests.
* [x] Documentation.

Jc2k:
This would be really handy for me.

I was thinking about writing a cert-manager plugin with kopf - something like [this](https://github.com/jetstack/cert-manager-external-issuer-example).

With this feature I would be able to filter out `CertificateRequests` that weren't "assigned" to my operator by looking at `spec.issuerRef`.

(Presumably these objects would then not be tagged with the kopf last-configuration annotations etc too).

nolar:
I think, it is easily implementable now. Since at least some filters were added in #123 (by labels/annotations), adding more filters is easy. 

A starting place to look for: https://github.com/zalando-incubator/kopf/blob/0.23.2/kopf/reactor/registries.py#L570-L623

Jc2k:
[...]

PR #258, opened 2019-11-27
Filter by arbitrary callback function

Jc2k:
# Filter by arbitrary callback function

> Issue : #98

## Description

In some cases, it might be needed that only a subset of resources are handled. An example might be implementing the cert-manager external issuer process where cert-manager creates a CertificateRequest CRD and waits for it to be acted on and marked as complete, and each issuer should filter on `spec.issuerRef` to find the requests it should service.

This PR enables that use case by allowing you to specify a `when` callback that is run before invoking a handler to filter out handlers that are not appropriate.

## Types of Changes

## Review
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/kopf/reactor/invocation.py
+++ b/kopf/reactor/invocation.py
@@ -41,7 +41,11 @@ def context(
         for var, token in reversed(tokens):
             var.reset(token)
 
-def get_invoke_arguments(*args: Any, cause: Optional[causation.BaseCause] = None, **kwargs: Any) -> Dict[str, Any]:
+def get_invoke_arguments(
+    *args: Any,
+    cause: Optional[causation.BaseCause] = None,
+    **kwargs: Any
+) -> Dict[str, Any]:
     """
     Expand kwargs dict with fields from the causation.
     """
--- a/kopf/reactor/invocation.py
+++ b/kopf/reactor/invocation.py
@@ -9,7 +9,7 @@ import asyncio
 import contextlib
 import contextvars
 import functools
-from typing import Optional, Any, Union, List, Iterable, Iterator, Tuple
+from typing import Optional, Any, Union, List, Iterable, Iterator, Tuple, Dict
 
 from kopf import config
 from kopf.reactor import causation
@@ -41,6 +41,51 @@ def context(
         for var, token in reversed(tokens):
             var.reset(token)
 
+def get_invoke_arguments(*args: Any, cause: Optional[causation.BaseCause] = None, **kwargs: Any) -> Dict[str, Any]:
+    """
+    Expand kwargs dict with fields from the causation.
+    """
+    new_kwargs = {}
+    new_kwargs.update(kwargs)
+
+    # Add aliases for the kwargs, directly linked to the body, or to the assumed defaults.
+    if isinstance(cause, causation.BaseCause):
+        new_kwargs.update(
+            cause=cause,
+            logger=cause.logger,
+        )
+    if isinstance(cause, causation.ActivityCause):
+        new_kwargs.update(
+            activity=cause.activity,
+        )
+    if isinstance(cause, causation.ResourceCause):
+        new_kwargs.update(
+            patch=cause.patch,
+            memo=cause.memo,
+            body=cause.body,
+            spec=dicts.DictView(cause.body, 'spec'),
+            meta=dicts.DictView(cause.body, 'metadata'),
+            status=dicts.DictView(cause.body, 'status'),
+            uid=cause.body.get('metadata', {}).get('uid'),
+            name=cause.body.get('metadata', {}).get('name'),
+            namespace=cause.body.get('metadata', {}).get('namespace'),
+        )
+    if isinstance(cause, causation.ResourceWatchingCause):
+        new_kwargs.update(
+            event=cause.raw,
+            type=cause.type,
+        )
+    if isinstance(cause, causation.ResourceChangingCause):
+        new_kwargs.update(
+            event=cause.reason,  # deprecated; kept for backward-compatibility
+            reason=cause.reason,
+            diff=cause.diff,
+            old=cause.old,
+            new=cause.new,
+        )
+
+    return new_kwargs
+
 
 async def invoke(
         fn: Invokable,
@@ -62,41 +107,7 @@ async def invoke(
     See: https://pymotw.com/3/asyncio/executors.html
     """
 
-    # Add aliases for the kwargs, directly linked to the body, or to the assumed defaults.
-    if isinstance(cause, causation.BaseCause):
-        kwargs.update(
-            cause=cause,
-            logger=cause.logger,
-        )
-    if isinstance(cause, causation.ActivityCause):
-        kwargs.update(
-            activity=cause.activity,
-        )
-    if isinstance(cause, causation.ResourceCause):
-        kwargs.update(
-            patch=cause.patch,
-            memo=cause.memo,
-            body=cause.body,
-            spec=dicts.DictView(cause.body, 'spec'),
-            meta=dicts.DictView(cause.body, 'metadata'),
-            status=dicts.DictView(cause.body, 'status'),
-            uid=cause.body.get('metadata', {}).get('uid'),
-            name=cause.body.get('metadata', {}).get('name'),
-            namespace=cause.body.get('metadata', {}).get('namespace'),
-        )
-    if isinstance(cause, causation.ResourceWatchingCause):
-        kwargs.update(
-            event=cause.raw,
-            type=cause.type,
-        )
-    if isinstance(cause, causation.ResourceChangingCause):
-        kwargs.update(
-            event=cause.reason,  # deprecated; kept for backward-compatibility
-            reason=cause.reason,
-            diff=cause.diff,
-            old=cause.old,
-            new=cause.new,
-        )
+    kwargs = get_invoke_arguments(*args, cause=cause, **kwargs)
 
     if is_async_fn(fn):
         result = await fn(*args, **kwargs)  # type: ignore
--- a/tests/registries/legacy/test_legacy_decorators.py
+++ b/tests/registries/legacy/test_legacy_decorators.py
@@ -189,7 +189,7 @@ def test_on_field_with_all_kwargs(mocker):
     cause = mocker.MagicMock(resource=resource, reason=Reason.UPDATE, diff=diff)
     mocker.patch('kopf.reactor.registries.match', return_value=True)
 
-    when = lambda body: False
+    when = lambda **_: False
 
     @kopf.on.field('group', 'version', 'plural', 'field.subfield',
                    id='id', timeout=123, registry=registry,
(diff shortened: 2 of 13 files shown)


Jc2k:
I couldn't use `reactor.invocation.invoke` as suggested in #98 because it's async and match is not.  This would mean making things like `iter_cause_handlers` async and felt like a bridge too far for my first PR. I don't think we should be doing I/O during the filter step anyway?

So I can see 2 ways forward:

 * The current approach in my PR - to only base body to the handler
 * Make a helper in `reactor.invocation` that does the args + kwargs setup and use it for the existing `invoke` method. Add an `invoke_blocking` and use that with my when function.

What do you think?

nolar:
True, it is async. The main purpose of this is for the [async handlers](https://kopf.readthedocs.io/en/latest/async/), which are executed directly in the stack of the operator, without thread executors. Both async and sync-threaded approaches are needed for remote I/O operation.

You are right: Unlike the handlers, the callbacks should not contain any I/O, so there is no reason for them to be async or thread-pool'ed.

I suggest to go the 2nd way: extract the args/kwargs generation from the current `invoke()` as a regular sync function, with `invoke()` just calling it and then doing whatever it does now for the function invocation. And then, you can call that callback with the kwargs generation function and a direct call — there is no need to use `invoke()` then (assuming it is always synchronous and non-blocking, as the callbacks should be).

This solution seems clean enough, and can be reused in some other places with sync-only callbacks. And it provides the same conventions on the kwargs, as any other invocables/callables in the framework (incl. the per-object `logger` kwarg & etc).

Jc2k:
I've hit a couple of snags while working on this.

The first makes me feel like we are violating our layering somewhat:

 * The invoke function renders `kwargs` based on a `cause` object
 * To get a cause we need to set `requires_finalizer`
 * To set `requires_finalizer` we need to call `match()` to see if any `DELETE` handlers apply to this resource
 * To call match we want to apply the new `when` filters
 * We want the new `when` filters to get the same kwargs as the handler would
 * That means it needs a `cause` object.

One option is to generate a dummy cause where `requires_finalizer` is `False` and when we have called our `match` functions make the real one... but that seems like a hack.

I'm leaning towards making some of the kwargs available but just the ones i can get from the body alone. In particular for requires_finalizer its weird to make whether or not there is a finalizer dependent on the cause.reason.

nolar:
Quite an interesting and non-trivial case of _the chicken & the egg_ problem.

Let me write down few thoughts on how it can be solved (though, I didn't check that in action — there can be some other problems in each option):

---

**Option A:**

To break the circle, I suggest that we think of some use-cases. Which information would be needed to these callbacks? Body/spec/meta/status — certainly. Logger — most likely. Memo — maybe. `diff`/`old`/`new` -- highly unlikely: in the context of the deletion, it is alwayys a diff from whatever to `None`.

Perhaps, we can always create and use a `ResourceWatchingCause` object, and also pass it to the callback invocation, while passing `ResourceChangingCause` for the handlers. Con: it requires passing two causes to the methods. It can be worked around by converting `ResourceChangingCause` to contain a `ResourceWatchingCause` rather than being a sibling class. Con: the structure of causes as contextual containers becomes complicated (hierarchical composition instead of a flat field list).

Or we can use the base class `ResourceCause` directly:

```python
async def resource_handler(...):
    ...

    resource_base_cause = causation.ResourceCause(
        logger=logger,
        resource=resource,
        patch=patch,
        body=event['object'],
        memo=memory.user_data,
    )

    if registry.has_resource_watching_handlers(resource=resource):
        resource_watching_cause = causation.detect_resource_watching_cause(
            ...
        )
        await handle_resource_watching_cause(cause=resource_watching_cause, ...)

    if registry.has_resource_changing_handlers(resource=resource):
        resource_changing_cause = causation.detect_resource_changing_cause(
            ...,
            requires_finalizer=registry.requires_finalizer(cause=resource_base_cause),  # <<<<< HERE
        )
        delay = await handle_resource_changing_cause(cause=resource_changing_cause, ...)
```

As a follow-up, the watching/changing cause detection can use the base cause as a source of fields instead of using all fields explicitly, but this can be left for future refactorings.

In the future, if the base class is declared in the involved function signatures, we will be able to replace the objects with the actual watching/changing clauses without breaking the typing system — in case we find some other way to solve this problem.

This will allow the deletion handlers to be filtered using the existing `invoke()` function, though will limit the kwargs available: there will be no `old`/`new`/`diff` or `initial` or raw `event` — but they are probably useless for filtering purposes anyway.

---

**Option B:**

The semantically weak link in this chain of dependencies seems to be the `requires_finalizer` flag. It is needed to detect whether the finalizer should be added on the object's first appearance (ACQUIRE/RELEASE pseudo-reasons), or should we go directly to the creation/deletion handlers (CREATE/UPDATE/DELETE real reasons). And this depends on the existence of the _matching_ mandatory deletion handlers. As such, it should not affect the filtering callback, in theory.

Can we redesign the solution so that the handler selection happens that way?

Can we detect a cause without passing `requires_finalizer` to `detect_resource_changing_cause()`? We can, but it comes at a cost: the following two conditions should be moved from `causation.detect_resource_changing_cause()` to `handling.handle_resource_changing_cause()`:

```python
    if requires_finalizer and not finalizers.has_finalizers(body):
        return ResourceChangingCause(reason=Reason.ACQUIRE, **kwargs)
    if not requires_finalizer and finalizers.has_finalizers(body):
        return ResourceChangingCause(reason=Reason.RELEASE, **kwargs)
```

Not only their execution should be there (as it is now already), but also their detection, which should happen strictly before actually detected cause's reasons are executed (i.e. before `if cause.reason in causation.HANDLER_REASONS`).

Con: we will "detect" the resource-changing causes, but will not use them in the ACQUIRE/RELEASE situations. But this does not produce any significant extra CPU overhead — happens once or twice per resource lifetime.

So, it will be something like that.

As is:

```python
async def handle_resource_changing_cause(...):
    if cause.reason in causation.HANDLER_REASONS:
        ...

    if cause.reason == causation.Reason.ACQUIRE:
        logger.debug("Adding the finalizer, thus preventing the actual deletion.")
        finalizers.append_finalizers(body=body, patch=patch)

    if cause.reason == causation.Reason.RELEASE:
        logger.debug("Removing the finalizer, as there are no handlers requiring it.")
        finalizers.remove_finalizers(body=body, patch=patch)
```

To be:

```python
async def handle_resource_changing_cause(...):

    requires_finalizer = registry.requires_finalizer(resource=cause.resource, body=cause.body)
    has_finalizer = finalizers.has_finalizers(body=cause.body)

    if requires_finalizer and not has_finalizer:
        logger.debug("Adding the finalizer, thus preventing the actual deletion.")
        finalizers.append_finalizers(body=body, patch=patch)
        return

    if not requires_finalizer and has_finalizer:
        logger.debug("Removing the finalizer, as there are no handlers requiring it.")
        finalizers.remove_finalizers(body=body, patch=patch)
        return

    if cause.reason in causation.HANDLER_REASONS:
        ...
```

In that case, the cause detection will not require `registry.requires_finalizer()`, and so the circular dependency will be broken, and so the callbacks will not be called during the cause detection, and so the solution becomes clean.

Do I miss any complications in that approach?

---

Both of the solutions seems relatively clean to me (from the code's semantics point of view).

I'm not sure which one is better. Are there any future hypothetical use-cases that we can consider to make one option more preferred than the other? E.g. different finalizers for different operators on the same resource kind (as in your case), so that they do not collide with each other?

Jc2k:
I implemented option B [here](https://github.com/zalando-incubator/kopf/pull/288). I will rebase on top of this branch and see if unblocks me.

Jc2k:
Ok, i have rebased this on top of https://github.com/zalando-incubator/kopf/pull/288 and all the tests are passing. Sorry it took me so long!

Jc2k:
Saw that you merged #288, so have rebased to try and make it a bit easier to review.

Jc2k:
@nolar I think this is ready for review again now.

kopf/reactor/invocation.py
```diff
@@ -41,39 +41,25 @@ def context(
...
             var.reset(token)
 
-
-async def invoke(
-        fn: Invokable,
-        *args: Any,
-        cause: Optional[causation.BaseCause] = None,
-        **kwargs: Any,
-) -> Any:
+def get_invoke_arguments(*args: Any, cause: Optional[causation.BaseCause] = None, **kwargs: Any) -> Dict[str, Any]:
```
nolar: Can you please put each arg here (and the resulting type) to its own line — to be consistent with the rest of the codebase?

tests/registries/legacy/test_legacy_decorators.py
```diff
@@ -174,10 +189,13 @@ def test_on_field_with_all_kwargs(mocker):
     cause = mocker.MagicMock(resource=resource, reason=Reason.UPDATE, diff=diff)
     mocker.patch('kopf.reactor.registries.match', return_value=True)
 
+    when = lambda body: False
```
nolar: Hm. Does it pass with no `**_`? If yes, that should be worrying, as it should fail here (if called at all).

tests/registries/test_decorators.py
```diff
@@ -257,12 +262,15 @@ def test_on_resume_with_all_kwargs(mocker, reason):
     cause = mocker.MagicMock(resource=resource, reason=reason, initial=True, deleted=False)
     mocker.patch('kopf.reactor.registries.match', return_value=True)
 
+    when = lambda body: False
```
nolar: Same here and below — with no `**_`.

tests/registries/legacy/test_legacy_decorators.py
Jc2k: It's technically incorrect and i've fixed it in my next push, but this is just asserting that the decorator propagates `when` to the handler so you could put an object with no __call__ there and it would still pass so i'm not worried that it passes.

tests/registries/test_decorators.py
Jc2k: Same as ^

kopf/reactor/invocation.py
Jc2k: Fixed in last push

tests/registries/legacy/test_legacy_decorators.py
nolar: Ah, I see. I missed that part. Thanks for clarifying.

nolar:
Awesome! I've played with this filter in different edge cases. Works fine!

```python
def flt(spec, **_):
    return spec.get('field') == 'valueZ'

@kopf.on.update('zalando.org', 'v1', 'kopfexamples', when=flt)
def update_it(logger, **kwargs):
    pass

@kopf.on.delete('zalando.org', 'v1', 'kopfexamples', when=flt)
def delete_it(logger, **kwargs):
    pass
```

With the deletion handlers, there is a fancy behaviour with addition/removal of a finalizer: if the filter function uses the spec/status fields of the object itself, and the object is edited/patched to satisfy the criterion, the finalizer is added; when edited/patched to miss it, the finalizer is removed — before executing the relevant handlers.

This can lead to some fancy effects if the filter functions sporadically fluctuate for the same object based on some external circumstances (e.g. `random.randint(0, 1)`), and there are no other mandatory deletion handlers, which would make the finalizer persistent. 

But let it be a problem of developers who design the app this way. 

This behaviour is fully correct (logically).

PR #266, opened 2019-12-07
Rename "cooldown" to "backoff" for handler retrying

nolar:
Fix a terminology issue, but keep it backward compatible.
 
> Issue : #222, #16

## Description

Due to my not so good English vocabulary, a slightly inappropriate word was used for the delay between retries of the handlers in #222: "cooldown". 

"Cooldowns" are used in the RPG games for the spells. In Computer Science, the proper word would be ["backoff"](https://en.wiktionary.org/wiki/backoff): 

* [exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff), 
* [`retry` library's `backoff=` kwarg](https://github.com/invl/retry#retry-decorator), 
* Kubernetes pods' `CrashLoopBackOff`, 
* Kubernetes jobs' [`spec.backoffLimit`](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#pod-backoff-failure-policy),
* etc.

It is better to fix this terminology issue before it spreads all over the codebase.  _(It is also possible that this kwarg will be extended to "exponential backoff" in the future.)_

The handlers'/decorators' `cooldown=` kwarg and `.cooldown` read-only property are kept there for backward compatibility as aliases for `backoff=`/`.backoff` — since there was at least one public release with this naming (we do not know if it is already used somewhere or not). 

The aliases are tested only to minimally guarantee that they work as expected until they are removed in the next major release (1.0). The rest of the tests are switched to the proper name.

## Types of Changes

- Refactor/improvements

## Review
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/docs/errors.rst
+++ b/docs/errors.rst
@@ -123,8 +123,8 @@ Once the number of retries is reached, the handler fails permanently.
 By default, there is no limit, so the retries continue forever.
 
 
-Cool-down on errors
-===================
+Backoff
+=======
 
 The interval between retries on arbitrary errors, when an external environment
 is supposed to recover and be able to succeed the handler execution,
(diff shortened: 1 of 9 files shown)


samurang87:
RPG's FTW

Issue #275, opened 2019-12-16
kopf crashes on large enough resources


pshchelo:
kopf 0.23.2 fails when a field in the resource being watched exceeds certain limit

in our case it is a secret with value size (in base64) of about 43KB.
Kubernetes itself accepts such Secret just fine, and Kubernetes docs state that the max secret size is 2MB.

## Expected Behavior
everything works

## Actual Behavior
kopf crashes and stops with the following error (for full trace see below in the repro)
`kopf.reactor.running [ERROR   ] Root task 'watcher of secrets' is failed: ValueError('Line is too long')`

it seems this has to do with the value of DEFAULT_LIMIT in the aiohttp streams
https://github.com/aio-libs/aiohttp/blob/6a5ab96bd9cb404b4abfd5160fe8f34a29d941e5/aiohttp/streams.py#L20

raising it to appropriate value solves the issue

It would be good to have possibility to pass custom value of limit to the `aiohttp.streams.StreamReader`:

https://github.com/aio-libs/aiohttp/blob/6a5ab96bd9cb404b4abfd5160fe8f34a29d941e5/aiohttp/streams.py#L109

or, if letting users to configure it in `kopf` is not an option, raise it by default to `2**20` so 2MB passes (translating to `2 ** 21` for `self._high_water` in the `StreamReader`)

## Specifications

  - Platform: Ubuntu 18.04
  - Kubernetes version: *1.15*
  - Python version: *3.7.3*
  - Python packages installed: relevant ones are *kopf 0.23.2* and *aiohttp 3.6.2*
  
## Steps to Reproduce the Problem

(using `kopf` namespace as example)

  1. Start a simple kopf handler that watches secrets:
```
$ cat secrets.py
import kopf

@kopf.on.create("", "v1", "secrets", labels={"foo": "bar"})
@kopf.on.update("", "v1", "secrets", labels={"foo": "bar"})
def apply(name, logger, event, **kwargs):
    logger.info(f"Event {event} for secret {name}")
    
$ kopf run --standalone -n kopf secrets.py
```

  2. Generate and create big enough secret
```
$ cat gen-big-secret.py

import base64
import random
import string
import yaml

SIZE = 2**15  # <-- WHILE BIG ENOUGH, NOT EVEN CLOSE TO 2MB AFTER BASE64 ENCODE

secret = {
    "kind": "Secret",
    "apiVersion": "v1",
    "type": "Opaque",
    "metadata": {
        "name": "too-big-secret",
[...]

Issue #272, opened 2019-12-16
CI/CD broken with Coverage 5.0


nolar:
## Expected Behavior

CI/CD works, especially when there are no unrelated code changes.

## Actual Behavior

A sample broken build: https://travis-ci.org/zalando-incubator/kopf/builds/625639579

```
Collecting coveralls (from -r requirements.txt (line 14))
  Downloading https://files.pythonhosted.org/packages/d7/d6/7d32c9c34a621c201c4bb3f61f4c828ee8655f2c7edc3a547e1dea6a7e67/coveralls-1.9.2-py2.py3-none-any.whl

Collecting coverage>=4.4 (from pytest-cov->-r requirements.txt (line 8))
  Downloading https://files.pythonhosted.org/packages/b6/d8/cb3f8bbae33db7f24027721d61f73145f25bfbcf078a23b2998478e5258c/coverage-5.0-cp37-cp37m-manylinux1_x86_64.whl (226kB)

coveralls 1.9.2 has requirement coverage<5.0,>=3.6, but you'll have coverage 5.0 which is incompatible.

Successfully installed PyYAML-5.2 aiohttp-3.6.2 aiojobs-0.2.2 aresponses-1.1.2 async-timeout-3.0.1 asynctest-0.13.0 chardet-3.0.4 click-7.0 codecov-2.0.15 coverage-5.0 coveralls-1.9.2 docopt-0.6.2 freezegun-0.3.12 idna-2.8 iso8601-0.1.12 kopf multidict-4.7.1 mypy-0.740 mypy-extensions-0.4.3 pykube-ng-19.10.0 pytest-aiohttp-0.3.0 pytest-asyncio-0.10.0 pytest-cov-2.8.1 pytest-mock-1.13.0 python-dateutil-2.8.1 requests-2.22.0 typed-ast-1.4.0 typing-extensions-3.7.4.1 urllib3-1.25.7 yarl-1.4.2
```

Coveralls has a restriction on `<5.0` ([link](https://github.com/coveralls-clients/coveralls-python/blob/1.9.2/setup.py#L36)), but pip does not respect it.

Since `coverage==5.0` has been released 2 days ago on 14.12.2019 ([link](https://github.com/nedbat/coveragepy/releases/tag/coverage-5.0)), it broke coveralls, and thus the Kopf's CI/CD builds.

Issue #269, opened 2019-12-12
CI/CD broken with Kubernetes 1.17.0 & Minikube 1.6.0 due to RBAC


nolar:
## Expected Behavior

CI/CD runs.

## Actual Behavior

CD/CD fails for Kubernetes "latest" version (1.17.0) with the latest Minikube 1.6.0. 

A sample broken build: https://travis-ci.org/zalando-incubator/kopf/builds/623752971

Caused by Minikube not able to start Kubernetes 1.17.0 with RBAC authentication. Described the issue in https://github.com/kubernetes/minikube/issues/6061

It is not Kopf-related, as it does not get to the tests yet.

PR #276, opened 2019-12-16
[275] Parse JSON lines ourselves to avoid aiohttp's "Line is too long"

nolar:
Parse the JSON lines of Kubernetes watch-streams with no memory limit.

> Issue : #275

## Description

Kubernetes can send JSON lines with huge objects. E.g., secrets of 2MB in #275. 

`aiohttp` has a hard-coded limit the lines received from the stream, and it is 128KB ([`aiohttp.streams.DEFAULT_LIMIT = 2**16`](https://github.com/aio-libs/aiohttp/blob/6a5ab96bd9cb404b4abfd5160fe8f34a29d941e5/aiohttp/streams.py#L20) = 64 KB for the low watermark of the buffer, multiplied by 2x for high watermark, where the exception is raised).

There is no way to externally configure the buffer size of the aiohttp's StreamReader, so as there is no way to inject our own `StreamReader` instances with `limit=` properly set. The [docs say](https://docs.aiohttp.org/en/stable/streams.html):

> User should never instantiate streams manually but use existing `aiohttp.web.Request.content` and `aiohttp.ClientResponse.content` properties for accessing raw BODY data.

For this reason, we implement our own per-line iterator with no memory control — it takes as much memory, as it is needed. Though, we try not to waste it too much by not having duplicates of each line in the buffer and by not having multiple copies of the same line.

The efficiency is not tested and is not measured. Specifically, it can be slow if there are hundreds or thousands of such huge resources or events are happening very often. But we assume it is sufficient for now, and can later be improved when some performance optimisations and measurements are performed.

_FYI: The provided test clearly shows how the pure aiohttp approach fails with "Line is too long" — if used without this wrapping iterator._

## Types of Changes

## Review
- [ ] Tests

### DIFF
--- a/kopf/clients/watching.py
+++ b/kopf/clients/watching.py
@@ -158,6 +158,41 @@ async def watch_objs(
     response.raise_for_status()
 
     async with response:
-        async for line in response.content:
+        async for line in _iter_lines(response.content):
             event = cast(bodies.RawEvent, json.loads(line.decode("utf-8")))
             yield event
+
+
+async def _iter_lines(
+        content: aiohttp.StreamReader,
+) -> AsyncIterator[bytes]:
+    """
+    Iterate line by line over the response's content.
+
+    Usage::
+
+        async for line in _iter_lines(response.content):
+            pass
+
+    This is an equivalent of::
+
+        async for line in response.content:
+            pass
+
+    Except that the aiohttp's line iteration fails if the accumulated buffer
+    length is above 2**17 bytes, i.e. 128 KB (`aiohttp.streams.DEFAULT_LIMIT`
+    for the buffer's low-watermark, multiplied by 2 for the high-watermark).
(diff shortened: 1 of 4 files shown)


pshchelo:
Thanks a lot for a quick patch!
I will test it tomorrow on our actual use case and report back.

nolar:
PS: Releasing it as 0.23.3 hotfix on top of 0.23.2 is not possible, as the CI/CD of that old state is broken due to new releases of the external components (#272 #269). So, this fix goes on top of the master — with [all the changes](https://github.com/zalando-incubator/kopf/compare/0.23.2...master) that are already there — and will be released as 0.24.

Preview of 0.24: 

image

pshchelo:
@nolar seems to work fine with us! eagerly waiting for 0.24 release now ...

Issue #281, opened 2020-01-02
timeout errors are logged at exception level


pshchelo:
## Expected Behavior
no traces in logs

## Actual Behavior
from time to time we see the following in the controller logs:
```
[2020-01-02 11:39:42,792] kopf.reactor.queuein [ERROR   ] functools.partial(<function resource_handler at 0x7fcd14aec950>, lifecycle=<function asap at 0x7fcd14add9e0>, registry=<kopf.toolkits.legacy_registries.SmartGlobalRegistry object at 0x7fcd1423a0d0>, memories=<kopf.structs.containers.ResourceMemories object at 0x7fcd1488ec90>, resource=Resource(group='', version='v1', plural='secrets'), event_queue=<Queue at 0x7fcd14322d50 maxsize=0 _getters[1] tasks=217>) failed with an exception. Ignoring the event.
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/kopf/reactor/queueing.py", line 179, in worker
    await handler(event=event, replenished=replenished)
  File "/usr/local/lib/python3.7/site-packages/kopf/reactor/handling.py", line 224, in resource_handler
    await patching.patch_obj(resource=resource, patch=patch, body=body)
  File "/usr/local/lib/python3.7/site-packages/kopf/clients/auth.py", line 46, in wrapper
    return await fn(*args, **kwargs, session=session)
  File "/usr/local/lib/python3.7/site-packages/kopf/clients/patching.py", line 54, in patch_obj
    raise_for_status=True,
  File "/usr/local/lib/python3.7/site-packages/aiohttp/client.py", line 504, in _request
    await resp.start(conn)
  File "/usr/local/lib/python3.7/site-packages/aiohttp/client_reqrep.py", line 860, in start
    self._continue = None
  File "/usr/local/lib/python3.7/site-packages/aiohttp/helpers.py", line 596, in __exit__
    raise asyncio.TimeoutError from None
concurrent.futures._base.TimeoutError
```

It seems it would be enough to handle `asyncio.TimeoutError` here
https://github.com/zalando-incubator/kopf/blob/d74a0638da01037d65e8b6c7b041c4f9c2e257cd/kopf/reactor/queueing.py#L178-L182
similar to how it is handled separately several times in the code before that,
and log only an error, not an exception (or may be even just a warning)

## Steps to Reproduce the Problem
[...]

PR #282, opened 2020-01-02
Do not log TimeoutError as exception

pshchelo:
use warning instead

issue #281

# One-line summary

Log asyncio.TimeoutError as warning instead of exception when running handler code

## Description
Log asyncio.TimeoutError as warning instead of exception when running handler code

## Types of Changes

## Review
- [ ] Tests
- [ ] Documentation


nolar:
So, there are 2 concerns regarding this PR:

1. 

Everything client-related is kept isolated inside `kopf.clients`. Everything outside of that package is completely unaware of which clients are used, and how does the framework communicate with the APIs.

Even if the client is raw `aiohttp`, it is still a client. Can be changed to e.g. `requests-async` or some other asyncio-based K8s-client later (no such plans as of now).

The errors are part of the clients. Therefore, the timeout errors handling should never leave `kopf.clients` package.

Beside, `asyncio.TimeoutError` is a generic timeout error for coroutines and tasks and locks and events and sleeps and whatever else in asyncio — not necessary related to the API communication via aiohttp. It can be a result of bad code solution or whatever else — and should be noticed as such. Python's Zen says: _"Errors should never pass silently."_ — and I agree here.

So, the error catch-and-suppress should be done directly at source. Is it only in patching?

---

2. 

Even when these errors are moved to `kopf.clients`, and are suppressed directly at source in patching/watching/other API calls, I have some doubts that this is a good approach. Why shouldn't the operator fail and restart if the environment is hostile and unstable?

There is already connection error suppressing in `kopf/clients/watching.py`:

```python
@auth.reauthenticated_stream
async def watch_objs():
    ………
    except (aiohttp.ClientConnectionError, aiohttp.ClientPayloadError):
        pass
```

But it is there based on an assumption that the watcher task restarts the watching queries once they are finished — and K8s API closes the streaming connection server-side quite often, so it is necessary there.

The patching is not retried though. We can add some resiliency to retry N times (N=? 3? 10?) before actually failing with escalation and restart. This will make operators slightly more stable in such unreliable environments or poor connectivity — instead of focusing on fixing the root cause itself (i.e. environment/connectivity).

There are few libraries for retrying — we can just use them. E.g. [retry](https://pypi.org/project/retry/) and [retrying](https://pypi.org/project/retrying/) — which one is better feature-wise?

---

Regarding the 1st point: the client-specific isolation to `kopf.clients` should be definitely done.

Regarding this 2nd point: making the API communication slightly more resilient sounds reasonable (especially since it is already done in some places implicitly) — I am still thinking on it.

nolar:
@samurang87 Still open indeed. Kopf shifted away from my work tasks, so I am now preparing some organisational things to fix that and move it to my personal sphere. Definitely need to make some cleanup with issues & PRs.

However, this specific PR had some unaddressed concerns mentioned in the comment above. And there are pending code changes which address the same aspects of the framework: 

* #368 for initial cluster connection errors (watching, not processing).
* #351 for throttling in case of processing errors (incl. API communication).

If there will be other issues left with cluster connectivity (to be investigated separately — there were a few other reports), they should be addressed in that codebase with these PRs merged.

So, I will close this PR for now. The issue remains open. Nevertheless, @pshchelo I am grateful for your contribution. Thank you!

PR #303, opened 2020-01-30
Update the Dockerfile to fix the handlers.py path inside container

prakashkl88:
When the Dockerfile mentioned here is copied as is, it does not run the container as the path inside the container is incorrect. Fixing that

## What do these changes do?

Reflects the correct Dockerfile doc to be used

## Description

Incorrect path in README.doc

Enhanced UX

N/A

End-users

N/A

No

## Type of changes

- Mostly documentation and examples (no code changes)

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ into a docker image with CI/CD tool of your preference.
 FROM python:3.7
 ADD . /src
 RUN pip install kopf
-CMD kopf run handlers.py
+CMD kopf run /src/handlers.py
 ```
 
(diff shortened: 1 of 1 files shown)


trulede:
Has anyone ever used this docker file? I tried today and found it did not work (missing kubernetes from the pip install list). However, it could just be caused by the way I build my Docker Images, which is from an Alpine base image.

If it is a problem, the missing "kubernetes" from _pip install_, I will happily make a PR with that change.

PR #307, opened 2020-02-05
Add kubernetes to the list of installed pip modules.

trulede:
## What do these changes do?

The example Dockerfile was missing a pip import for the kubernetes module.

## Description

People following the documentation have a greater chance of success at their first attempt to use Kopf.

## Issues/PRs

 

> Issues: 

> Related:

## Type of changes

- Mostly documentation and examples (no code changes)

## Checklist

- [ ] I think the code is well written


nolar:
@trulede Can you please point out where `kubernetes` is used or linked in README?

One of the main points is that Kopf is client-agnostic. Operator developers can use `kubernetes`, `pykube-ng`, or whatever they want (if they need any client at all). 

I would better give no preferences to any one of the available clients, and would not highlight them in any other way, especially on the main introduction page — unless it breaks some examples and/or the quick-starting guides (though, I believe, they have their own `pip install ...` lines in the docs — in that case, those lines should be fixed instead).

trulede:
Right. I see how your perspective can be different from mine.

Thanks.

nolar:
@trulede It might be the case with some special Kubernetes providers with complicated authentication schemas, like GCP/GKE, or maybe AWS/EKS. 

`kubernetes` contains some code to retrieve the tokens from their central authentication system. Kopf piggybacks on that code _if_ `kubernetes` is installed (same for `pykube-ng`), or falls back to plain simple authentication methdos otherwise.

It is a good idea to describe this case as something like "if you use GKE or EKS and require complicated authentication schemas, it is also recommended to `pip install kubernetes`" (the key word is _"recommended"_, not _"required"_) — both in README and in the docs. That would probably solve some issues of users who only start with Kopf in a new virtualenv.

Issue #204, opened 2019-10-09
Opearator handler shows connection rest by peer or sometimes just hangs and doesn't respond


amolkavitkar:
## Expected Behavior
The pod should respond immediately 

## Actual Behavior
- When operator pod is long standing, and if we send and create request for custom resource it just hangs and doesn't respond, we have to restart pod then it comes back alive

- Connection reset by peer in middle of creating resources

## Steps to Reproduce the Problem

  1. Create an operator pod and keep it long running for few days
  2. try to create custom resource it doesn't go through

Note (sometimes it gives connection reset by peer)
## Specifications

  - Platform: Ubuntu 16.04
  - Kubernetes version: *(use `kubectl version`)* 1.15
  - Python version: *(use `python --version`)* 3.5.2
  - Python packages installed: *(use `pip freeze --all`)*

```
```
connection reset by peer traceback
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/kopf/reactor/handling.py", line 387, in _execute
    lifecycle=lifecycle,  # just a default for the sub-handlers, not used directly.
  File "/usr/local/lib/python3.7/site-packages/kopf/reactor/handling.py", line 478, in _call_handler
    **kwargs,
  File "/usr/local/lib/python3.7/site-packages/kopf/reactor/invocation.py", line 79, in invoke
    result = await loop.run_in_executor(config.WorkersConfig.get_syn_executor(), real_fn)
  File "/usr/local/lib/python3.7/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/handlers.py", line 159, in create_fn
    obj = v1_client.create_namespaced_service(namespace, SVC_NOSELECTOR_TEMPLATE)
  File "/usr/local/lib/python3.7/site-packages/kubernetes/client/apis/core_v1_api.py", line 6934, in create_namespaced_service
    (data) = self.create_namespaced_service_with_http_info(namespace, body, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/kubernetes/client/apis/core_v1_api.py", line 7025, in create_namespaced_service_with_http_info
    collection_formats=collection_formats)
  File "/usr/local/lib/python3.7/site-packages/kubernetes/client/api_client.py", line 334, in call_api
    _return_http_data_only, collection_formats, _preload_content, _request_timeout)
[...]

PR #322, opened 2020-03-06
Allow for custom timeout in watch request

pshchelo:
## What do these changes do?
allow to set a custom timeout when making the `GET ..?watch=true` request to Kubernetes API if needed.

## Description

Currently the timeout is always None, which we find to be a reason for issues like #204 .
With this patch controller authors may override the timeout passed by kopf to the underlying aiohttp session request by doing something like this in their controller code:
```
import kopf

kopf.config.WatchersConfig.session_timeout = 10.0

@kopf.on(..)
def handler(*):
    ....
```
The default behavior is not changed.

## Issues/PRs

 

Issues: #204 

## Type of changes

## Checklist

- [ ] I think the code is well written

### DIFF
--- a/kopf/clients/watching.py
+++ b/kopf/clients/watching.py
@@ -201,7 +201,7 @@ async def watch_objs(
     # Talk to the API and initiate a streaming response.
     response = await context.session.get(
         url=resource.get_url(server=context.server, namespace=namespace, params=params),
-        timeout=aiohttp.ClientTimeout(total=None),
+        timeout=aiohttp.ClientTimeout(total=config.WatchersConfig.session_timeout),
     )
     response.raise_for_status()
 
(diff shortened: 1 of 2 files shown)


pshchelo:
interestingly enough, coveralls job fails and reports as not covered lines I did not change/add...

nolar:
Coveralls decrease is all fine. It fluctuates a little bit here & there.

Regarding the feature: Can you please refactor it to make the timeout as `float` instead of `aiohttp.ClientTimeout`? 

Usage of `aiohttp` is a hidden detail of the current implementation and can change in the future. The session/client timeouts will probably stay, just differently implemented inside of `kopf.clients.*`.

nolar:
UPD: Actually, `Optional[float] = None`, not just `float`.

pshchelo:
Ack, sounds reasonable, will do

Issue #19, opened 2019-04-02
Call handlers by time


nolar:
## Expected Behavior

Some of the handlers must be called regularly on the schedule (e.g. cron-like).

For example, to monitor the actual state of the system (as changed out of scope of the operator) with the declared state (as in the yaml files).

## Actual Behavior

Only the objects changes cause the handlers execution. 

The objects are not always changed when something else (unmonitored) happens in the cluster.

nolar:
Docs: 

* https://kopf.readthedocs.io/en/latest/daemons/
* https://kopf.readthedocs.io/en/latest/timers/

PR #329, opened 2020-03-12
Split handers & callbacks types for resource-related handlers

nolar:
## What do these changes do?

Move the code around and rename some types & classes for callbacks & handlers.
No user-facing changes.

## Description

This is the last preparing PR before the daemons & timers PR (#330).

Here, we just move the code around and rename classes. It would be messy to have these in the daemons & timers PR, so it goes separately.

Specifically, the resource-related handlers are split into resource-changing & resource-watching — same as it is already done for registries. These handlers & callbacks are de factor different anyway, but were declared as the same for laziness reasons. In the daemons & timers PR, also resource-spawning handlers & daemon/timer callbacks will be added with their own signatures.

No behavioural changes, no user-facing changes.

## Issues/PRs

> Issues: #19 

## Type of changes

- Refactoring (non-breaking change which does not alter the behaviour)

## Checklist

### DIFF
--- a/kopf/on.py
+++ b/kopf/on.py
@@ -25,8 +25,9 @@ from kopf.structs import dicts
 from kopf.structs import filters
 from kopf.structs import resources
 
-ResourceHandlerDecorator = Callable[[callbacks.ResourceHandlerFn], callbacks.ResourceHandlerFn]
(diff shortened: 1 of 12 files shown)


nolar:
@haikoschol Ouch. One more time plz (with a new commit)?

haikoschol:
> @haikoschol Ouch. One more time plz (with a new commit)?

Sorry about the delay. I saw it yesterday or so but then forgot about it again.

Issue #317, opened 2020-02-26
Question: What is the best approach to monitor children objects within their parent?


xocoatzin:
Hi all, 

I'm currently working on porting some code from [metacontroller](https://github.com/GoogleCloudPlatform/metacontroller) into Kopf. 

Metacontroller, gives you the option to receive callbacks when the monitored object changes, or any of its children is updated. 

For example, if I'm watching an object of kind `multijob`, which creates an arbitrary number of standard kubernetes `job`s, I would receive a callback if the children fail, restart, succeed, etc, which I can use to update the `status` field in the parent. 

I haven't been able to find a clean way to do the same thing in Kopf, other than adding separate listeners for both the parent/children, and within the children listeners update the parent CRD. Of course the example here is simplified, the actual application would have many dependencies and larger hierarchies of objects, and having this kind of inter-dependencies between listeners make them harder to maintain. 

Is there any better way to do this? Or is there any feature in Kopf that would make the management of children easier/cleaner?

Thanks!

nolar:
Related: #58 #264 
See also: https://github.com/zalando-incubator/kopf/issues/264#issuecomment-562845724

You are right, the only way is —as you said— _"adding separate listeners for both the parent/children, and within the children listeners update the parent CRD"_.

Keep in mind, that Kopf keeps one and only one watch-query (an API request) per resource kind no matter how many handlers are there for that resource kind. So, there should be no problems with APIs.

There is no simpler (i.e. few-liner) solution **at the moment**. 

A better solution is planned though — but rather later than sooner (because: priorities; and my regular employment takes time). 

Under the hood, it will be working exactly the same way, just with better DSL for handlers. Some ideation was happening in this [gist](https://gist.github.com/nolar/8c6233778a30a32fafd7f8d3a55a2cb4).

---

For all those looking for a solution/pattern and coming to this issue — here is an example, which we currently use for ourselves:

* _Label_ the children resources with name & namespace of the parent object (assuming they can be in different namespaces; if it is the same namespace by design, only the name is needed).

* _Watch_ for children resources that have this label (any value). In the watcher, get the name of the parent resource and patch its status field (e.g. `status.subpods`) with the status of the watched children resource (selected or agregated).

* Back in the parent resource, react to changes in that status field, and do the calculation on all the children overall statuses.

A sample skeleton code:

```python
# pip install kopf pykube-ng PyYAML
import kopf
import pykube
import yaml


[...]

Issue #271, opened 2019-12-15
Question: Run handler based on timer


arthurh4:
Hello, very nice project ! I would like to run a function to run to some checks based on a timer (eg: every 30 seconds) is there any existing handler I could use ? Thanks

nolar:
Hello. Thanks.

I'm currently working on this for #19, #150 and related issues, which will make it possible and easier. A draft is already functional, but I was slightly "distracted" by sickness.

Meanwhile, you can use this pattern:

```python
import asyncio
import kopf

@kopf.on.create('zalando.org', 'v1', 'kopfexamples')
@kopf.on.resume('zalando.org', 'v1', 'kopfexamples')
def spawn_the_monitor(namespace, name, memo, logger, **kwargs):
    if not hasattr(memo, 'monitor'):
        memo.monitor = asyncio.create_task(monitor_me(namespace, name, logger))

@kopf.on.delete('zalando.org', 'v1', 'kopfexamples')
def stop_the_monitor(memo, **kwargs):
    if hasattr(memo, 'monitor'):
        memo.monitor.cancel()

async def monitor_me(namespace, name, logger):
    while True:
        logger.info('Checking...')
        # check it here (BEWARE: sync calls block the whole operator)
        await asyncio.sleep(30)
```

The same would be with the `threading.Thread()`, except that it is harder to terminate — needs an event-flag to indicate that it should stop.

nolar:
Daemons and timers are added in `kopf>=0.27rc1`.

Release notes:

* https://github.com/zalando-incubator/kopf/releases/tag/0.27rc1

Docs:

* https://kopf.readthedocs.io/en/latest/daemons/
* https://kopf.readthedocs.io/en/latest/timers/

**Beware:** the changes are quite massive, and the daemons/timers are a new feature, they are not tested thoroughly yet, can behave weird. Either test carefully in an isolated cluster/namespace, or wait for some time until the feature is tested by others (e.g. by us).

In your case, it would be:

```python
import kopf

@kopf.timer(..., interval=30)
def fn(**kwargs):
    pass
```

For now, I close this issue as it is _presumably_ solved. Any bugs or misbehaviour or misunderstanding should be reported separately.

Issue #122, opened 2019-06-28
How to periodly poll something?


Kiddinglife:
Hi folks,
for example,
I need my controller to periodly poll the premetheus server for some metrics so that my contraoller can automatically scale up more pods.
thanks.

nolar:
@Kiddinglife Sorry, I didn't get the part with replicasets.

But for polling something in Kopf you can use the `@kopf.on.resume` handlers to start a background thread or an asyncio task.

It is better if resuming is used on combination with `@kopf.on.create`: either the object was created while the operator is up (creation), or the operator was started while the object exists (resuming) — it is usually either-either.

```python
@kopf.on.resume('zalando.org', 'v1', 'kopfexamples')
@kopf.on.create('zalando.org', 'v1', 'kopfexamples')
def start_poller(**kwargs):
    threading.Thread(target=poller, kwargs=kwargs).start()

def poller(**kwargs):
    while shouldstop:
        check_it()
        change_it()
        time.sleep(60)
```

Thread deduplication can be additionally secured with `uid` kwarg and by keeping a `dict{uid=>thread}` structure in the module:

```python
threads = {}

@kopf.on.resume('zalando.org', 'v1', 'kopfexamples')
@kopf.on.create('zalando.org', 'v1', 'kopfexamples')
def start_poller(**kwargs):
    uid = kwargs['uid']
    if uid not in threads or not threads[uid].isAlive():
        shouldstop = threading.Event()
        thread = threading.Thread(target=poller, args=(shouldstop,), kwargs=kwargs)
        threads[uid] = (thread, shouldstop)
        thread.start()
    else:
        pass  # already running

@kopf.on.delete('zalando.org', 'v1', 'kopfexamples')
def stop_polling(uid, **kwargs):
    thread, shouldstop = threads[uid]
    shouldstop.set()
    thread.join()  # can take ~60 seconds
    del threads[uid]

def poller(shouldstop, **kwargs):
    while not shouldstop.is_set():
        ......
        time.sleep(60)
```

The docs: 

* https://kopf.readthedocs.io/en/latest/handlers/?highlight=resume#cause-handlers

eshepelyuk:
@nolar is there any advantage of using dedicated threads for polling ? Versus using asyncio based libraries, like https://github.com/gawel/aiocron ?

nolar:
@eshepelyuk Cron-based functions can be sufficient for some cases. But they are unaware of Kubernetes resources in any form. 

[...]

PR #330, opened 2020-03-12
Daemons & timers for background processing and reconciliation

### LINKED ISSUE #19: Call handlers by time

nolar:
## What do these changes do?

Add background daemons & timers for resources.

## Description

Previously, the handlers were event-driven — i.e. when something changes on the resource. If nothing changes, no handlers are called. To have reconciliation or other activities happening during the lifetime of a resource, the developers had to do their own task/thread orchestration.

With this PR, the task/thread orchestration is generalised to a new framework feature to perform background reconciliation with external systems, locally managed applications, or even local children K8s resources not the event-driven way:

```python
import kopf

@kopf.daemon('zalando.org', 'v1', 'kopfexamples', initial_backoff=5.0, cancellation_backoff=10.0)
async def background_async(spec, logger, **_):
    while True:  # async handlers will be cancelled: instantly or with a graceful delay.
        logger.info(f"Ping from an async daemon: field={spec['field']!r}")
        await asyncio.sleep(2.5)

@kopf.timer('zalando.org', 'v1', 'kopfexamples', idle=15, interval=2, sharp=True)
def every_few_seconds_sync(spec, logger, **_):
    logger.info(f"Ping from a sync timer: field={spec['field']!r}")
```

The daemons run as long as the resource exists. They are stopped/cancelled when the resource is deleted (or when the operator exits/restarts). The daemons can be automatically restarted or left dead if exited (`restart_backoff`).

The timers are also triggered as long as the resource exists, but by schedule. They can be postponed by few seconds since the last change (`idle`), or since the resource creation / operator startup (`initial_backoff`), can have sharp or soft intervals (see the docs), etc.

For more examples & options, see `/examples/` folder and the doc updates in this PR/branch.

## Issues/PRs

> Issues: closes #19 

> Related: #150 #271 #317 #122 

## Type of changes

## Checklist

---
**WARNING:** This PR is a preview. It is runnable and usable, but few minor things are not finished yet:

* [ ] Tests
  * [ ] For the `stopper` synchronization primitive.
  * [ ] For daemon existing reason signalling (`DaemonStopperReason`).
  * [x] For daemon & timer decorators.
  * [x] For async daemon cancellation.
  * [ ] For sync daemon cancellation.
  * [ ] For timer guarding task cancellation.
* [ ] A sync-daemon did not exit on operator stopping in a try-run (it did exit before, something is broken? — check & fix.).

### DIFF
--- a/docs/daemons.rst
+++ b/docs/daemons.rst
@@ -211,7 +211,7 @@ It is possible to postpone the daemon spawning:
     import asyncio
     import kopf
 
-    @kopf.daemon('zalando.org', 'v1', 'kopfexamples', initial_backoff=30)
+    @kopf.daemon('zalando.org', 'v1', 'kopfexamples', initial_delay=30)
     async def monitor_kex(stopped, **kwargs):
         while True:
             await asyncio.sleep(1.0)
--- a/docs/daemons.rst
+++ b/docs/daemons.rst
@@ -1,0 +1,374 @@
+=======
+Daemons
+=======
+
+Daemons are a special type of handlers for background logic that accompanies
+the Kubernetes resources during their life cycle.
+
+Unlike event-driven short-running handlers declared with ``@kopf.on``,
+daemons are started when an object is created or an operator is started,
+and are capable of running indefinitely (or infinitely) long.
+
+The daemons are stopped when the object is deleted
+or the whole operator is exiting/restarting.
+
+
+Spawning
+========
+
+To have a daemon accompanying a resource of some kind, decorate a function
+with ``@kopf.daemon`` and make it run for long time or forever:
+
+.. code-block:: python
+
+    import asyncio
+    import time
+    import kopf
+
+    @kopf.daemon('zalando.org', 'v1', 'kopfexamples')
+    async def monitor_kex_async(**kwargs):
+        while True:
+            ...  # check something
+            await asyncio.sleep(10)
+
+    @kopf.daemon('zalando.org', 'v1', 'kopfexamples')
+    def monitor_kex_sync(stopped, **kwargs):
+        while not stopped:
+            ...  # check something
+            time.sleep(10)
+
+Synchronous functions are executed in threads, asynchronous functions are
+executed directly in the asyncio event loop of the operator -- same as with
+regular handlers, except that a separate thread pool is used for daemons
+to prevent thread pool depletion for regular handlers.
+See :doc:`async`.
+
+
+Termination
+===========
+
+The daemons are terminated when either their resource is marked for deletion,
+or the operator itself is exiting.
+
+In both cases, the daemons are requested to terminate gracefully by setting
+the :kwarg:`stopped` kwarg. The synchronous daemons MUST_, and asynchronous
+daemons SHOULD_ check for the value of this flag as often as possible:
+
+.. code-block:: python
+
+    import asyncio
+    import kopf
+
+    @kopf.daemon('zalando.org', 'v1', 'kopfexamples')
+    def monitor_kex(stopped, **kwargs):
+        while not stopped:
+            time.sleep(1.0)
+        print("We are done. Bye.")
+
+The asynchronous daemons can skip these checks if they define the cancellation
+timeout (see below). In that case, they can expect an `asyncio.CancelledError`
+to be raised at any point of their code (specifically, at any ``await`` clause):
+
+.. code-block:: python
+
+    import asyncio
+    import kopf
+
+    @kopf.daemon('zalando.org', 'v1', 'kopfexamples', cancellation_timeout=1.0)
+    async def monitor_kex(**kwargs):
+        try:
+            while True:
+                await asyncio.sleep(10)
+        except asyncio.CancelledError:
+            print("We are done. Bye.")
+
+With no cancellation timeout set, cancellation is not performed at all,
+as it is unclear for how long should the coroutine be awaited. However,
+it is cancelled when the operator exits and stops all "hung" left-over tasks
+(not specifically daemons).
+
+.. note::
+
+    The MUST_ / SHOULD_ separation is due to Python has no way of terminating
+    a thread unless the thread exits by its own. The :kwarg:`stopped` flag
+    is a way to signal the thread it should exit. If :kwarg:`stopped` is not
+    checked, the synchronous daemons will run forever or until an error happens.
+
+.. _MUST: https://tools.ietf.org/rfc/rfc2119.txt
+.. _SHOULD: https://tools.ietf.org/rfc/rfc2119.txt
+
+
+Timeouts
+========
+
+The termination sequence parameters can be controlled when declaring a daemon:
+
+.. code-block:: python
+
+    import asyncio
+    import kopf
+
+    @kopf.daemon('zalando.org', 'v1', 'kopfexamples',
+                 cancellation_backoff=1.0, cancellation_timeout=3.0)
+    async def monitor_kex(stopped, **kwargs):
+        while not stopped:
+            await asyncio.sleep(1)
+
+There are three stages how the daemon is terminated:
+
+* 1. Graceful termination:
+  * ``stopped`` is set immediately (unconditionally).
+  * ``cancellation_backoff`` is awaited (if set).
+* 2. Forced termination -- only if ``cancellation_timeout`` is set:
+  * `asyncio.CancelledError` is raised (for async daemons only).
+  * ``cancellation_timeout`` is awaited (if set).
+* 3a. Giving up and abandoning -- only if ``cancellation_timeout`` is set:
+  * A `ResourceWarning` is issued for potential OS resource leaks.
+  * The finalizer is removed, and the object is released for potential deletion.
+* 3b. Forever polling -- only if ``cancellation_timeout`` is not set:
+  * The daemon awaiting continues forever, logging from time to time.
+  * The finalizer is not removed and the object remains blocked from deletion.
+
+The ``cancellation_timeout`` is measured from the point when the daemon
+is cancelled (forced termination begins), not from where the termination
+itself begins; i.e., since the moment when the cancellation backoff is over.
+The total termination time is ``cancellation_backoff + cancellation_timeout``.
+
+.. warning::
+
+    When the operator is exiting, it has its own timeout of 5 seconds
+    for all "hung" tasks. This includes the daemons after they are requested
+    to exit gracefully and all timeouts are reached.
+
+    If the daemon termination takes longer than this for any reason,
+    the daemon will be cancelled (by the operator, not by the daemon guard)
+    regardless of the graceful timeout of the daemon. If this does not help,
+    the operator will be waiting for all hung tasks until SIGKILL'ed.
+
+.. warning::
+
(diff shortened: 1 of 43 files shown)


kopf/reactor/daemons.py
```diff
@@ -0,0 +1,394 @@
...
+        if daemon.task.done():
+            pass
+
+        elif backoff is not None and age < backoff:
+            if not stopper.is_set(reason=primitives.DaemonStoppingReason.DAEMON_SIGNALLED):
+                stopper.set(reason=primitives.DaemonStoppingReason.DAEMON_SIGNALLED)
+                logger.debug(f"Daemon {daemon_id!r} is signalled to exit gracefully.")
+            delays.append(backoff - age)
+
+        elif timeout is not None and age < timeout + (backoff or 0):
```
michaelnarodovitch: should it be rather `timeout - (backoff or 0)` here ?

kopf/reactor/daemons.py
```diff
@@ -0,0 +1,394 @@
...
+        backoff = daemon.handler.cancellation_backoff
+        timeout = daemon.handler.cancellation_timeout
+    elif isinstance(daemon.handler, handlers_.ResourceTimerHandler):
+        backoff = None
+        timeout = None
+    else:
+        raise RuntimeError(f"Unsupported daemon handler: {daemon.handler!r}")
+
+    # Whatever happens with other flags & logs & timings, this flag must be surely set.
+    daemon.stopper.set(reason=primitives.DaemonStoppingReason.OPERATOR_EXITING)
```
michaelnarodovitch: maybe something like...
```
        if not daemon.task.done():
            daemon.stopper.set(reason=primitives.DaemonStoppingReason.OPERATOR_EXITING)
            daemon.logger.debug(f"Daemon {daemon_id!r} is signalled to exit gracefully.")
```
? Otherwise there will be no log message if the daemon exits instantly.

nolar:
The tests are rudimentary, only for the smoke-testing of daemons & timers.  This PR took already too much time, and is too big.

The low-level primitives are tested indirectly through the existing tests, and manually by running examples with Minikube (including the examples 14 & 15 for daemons & timers).

The framework itself will be tested for backward compatibility by deployed a release candidate to our own clusters and utilising the new features (incl. daemons & timers).

More detailed tests for all little aspects will be added later as separate PRs — with lower priority (time-wise).

kopf/reactor/daemons.py
nolar: Indeed. Fixed.

nolar: It is correct as it is. Here, the `age` is measured since the termination itself has started. However, the `cancellation_timeout` should be measured since `task.cancel()`. Which means, the real age limit is `backoff` + `cancellation_timeout` — unlike the waiting timeout below for sleep-or-wait.

kopf/reactor/daemons.py
nolar: Right. I will take a look at this.

docs/daemons.rst
```diff
@@ -0,0 +1,374 @@
+=======
+Daemons
+=======
+
+Daemons are a special type of handlers for background logic that accompanies
```
haikoschol: "a special type of handler"

docs/daemons.rst
```diff
@@ -0,0 +1,374 @@
...
+the Kubernetes resources during their life cycle.
+
+Unlike event-driven short-running handlers declared with ``@kopf.on``,
+daemons are started when an object is created or an operator is started,
```
haikoschol: This is a bit ambiguous: Is one task created per resource that the daemon triggers on or is only ever one task started when the first resource of that kind is created and stopped when the last resource of that kind is deleted?

docs/daemons.rst
```diff
@@ -0,0 +1,374 @@
...
+            print("We are done. Bye.")
+
+With no cancellation timeout set, cancellation is not performed at all,
+as it is unclear for how long should the coroutine be awaited. However,
+it is cancelled when the operator exits and stops all "hung" left-over tasks
+(not specifically daemons).
+
+.. note::
+
+    The MUST_ / SHOULD_ separation is due to Python has no way of terminating
```
haikoschol: "due to Python having no way to" or "due to the fact that Python has no way to"

docs/daemons.rst
```diff
@@ -0,0 +1,374 @@
...
+.. code-block:: python
+
+    import kopf
+
+    @kopf.daemon('zalando.org', 'v1', 'kopfexamples')
+    async def monitor_kex(stopped, **kwargs):
+        while not stopped:
+            await stopped.wait(10)
+
+This way, the daemon will exit as soon as possible when the :kwarg:`stopped``
```
haikoschol: extra backtick at the end of :kwarg:`stopped``

docs/daemons.rst
```diff
@@ -0,0 +1,374 @@
...
+to the operator: i.e. on resource creation or operator startup.
+
+It is possible to postpone the daemon spawning:
+
+.. code-block:: python
+
+    import asyncio
+    import kopf
+
+    @kopf.daemon('zalando.org', 'v1', 'kopfexamples', initial_backoff=30)
```
haikoschol: `initial_backoff` might be a bit confusing since nothing is being backed off from yet, i.e. no attempt to perform an action has been made. Maybe `initial_delay`?

docs/daemons.rst
```diff
@@ -0,0 +1,374 @@
...
+    async def monitor_kex(stopped, **kwargs):
+        while True:
+            await asyncio.sleep(1.0)
+
+
+The daemon will not be started during the specified time after it should
```
haikoschol: This sentence is *definitely* confusing. :) Maybe: "The start of the daemon will be delayed by 30 seconds."

nolar: Paraphrased to this:

> The start of the daemon will be delayed by 30 seconds after the resource
> creation (or operator startup). For example, this can be used to give some time
> for regular event-driven handlers to finish without producing too much activity.

docs/daemons.rst
nolar: I probably misunderstood the meaning of this word ("back-off").

Renamed.

docs/daemons.rst
nolar: Fixed.

docs/daemons.rst
nolar: Fixed.

docs/daemons.rst
nolar: Paraphrased to this:

> Unlike event-driven short-running handlers declared with ``@kopf.on``,
> daemons are started for every individual object when it is created
> (or when an operator is started/restarted while the object exists),
> and are capable of running indefinitely (or infinitely) long.

Not sure if it is better though…

docs/daemons.rst
nolar: Fixed.

docs/timers.rst
```diff
@@ -0,0 +1,228 @@
...
+The default behaviour is to retry arbitrary error
+(similar to the regular resource handlers).
+
+When an error happens, its delay overrides the timer's schedule or life cycle:
+
+* For arbitrary exceptions, the timer's ``backoff=...`` option is used.
+* For `kopf.TemporaryError`, the error's ``delay=...`` option is used.
+* For `kopf.PermanentError`, the timer stops forever and is never retried.
+
+The timer's own interval is only used of the function exits successfully.
```
haikoschol: typo: `"if the function exits"

docs/daemons.rst
haikoschol: It is, thanks.

kopf/reactor/daemons.py
```diff
@@ -0,0 +1,435 @@
...
+) -> Collection[float]:
+    """
+    Ensure that all daemons are spawned for this individual resource.
+
+    This function can be called multiple times on multiple handling cycles
+    (though usually should be called on the first-seen occasion), so it must
+    be idempotent: not having duplicating side-effects on multiple calls.
+    """
+    if memory.live_fresh_body is None:  # for type-checking; "not None" is ensured in processing.
+        raise RuntimeError("A daemon is spawned with None as body. This is a bug. Please repor.")
```
haikoschol: typo: report

kopf/reactor/processing.py
```diff
@@ -56,75 +61,149 @@
...
+    # But not when it has already gone.
+    if deletion_is_ongoing and deletion_is_blocked \
+            and not resource_spawning_delays \
+            and not resource_changing_delays:
+        logger.debug("Removing the finalizer, thus allowing the actual deletion.")
+        finalizers.allow_deletion(body=body, patch=patch)
+
+    # Whatever was done, apply the accumulated changes to the object, or sleep-n-touch for delays.
     # But only once, to reduce the number of API calls and the generated irrelevant events.
+    # And only of the object is at least supposed to exist (not "GONE"), even if actually does not.
```
haikoschol: typo: "only *if* the object is"

kopf/reactor/processing.py
```diff
@@ -185,36 +266,72 @@
...
+        cause: causation.ResourceSpawningCause,
+) -> Collection[float]:
+    """
+    Spawn/kill all the background tasks of a resource.
+
+    The spawning and killing happens in parallel with the resource-changing
+    handlers invocation (even if it takes few cycles). For this, the signal
+    to terminate is sent to the daemons immediately, but the actual check
+    of their shutdown is performed only when all the on-deletion handlers
+    are succeeded (or after they were invoked if they are optional;
```
haikoschol: "have succeeded"

tests/handling/daemons/test_daemon_errors.py
```diff
@@ -0,0 +1,181 @@
...
+    event_object = {'metadata': {'finalizers': [FINALIZER]}}
+    await simulate_cycle(event_object)
+
+    await dummy.steps['called'].wait()
+    await dummy.wait_for_daemon_done()
+
+    assert dummy.mock.call_count == 1
+    assert k8s_mocked.patch_obj.call_count == 0
+    assert k8s_mocked.sleep_or_wait.call_count == 1  # one for each retry
+    assert k8s_mocked.sleep_or_wait.call_args_list[0][0][0] is None
```
michaelnarodovitch, line 29: This is not readable. Maybe `from unittest.mock import call` and
```
assert ...call_args_list == [call(*args, **kwargs)]
```
is better. Applies to the other stuff also. If tests breaks, then there might be obscure out of index exceptions.

nolar: Breaking the test in that case is intended. The first index is "secured" by a preceding check for the number of calls.

But you are right, making tests more readable is a good idea — but should be done for all the tests. Now, the tests in overall are in a clumsy state, grown over the year, and they were never refactored. Some of them are duplicating each other, some aspects of the code are just missing.

I've created #338 to collect the ideas on the tests, and added this one there.

PR #192, opened 2019-09-25
Obey the configured levels when posting from `logging`

nolar:
Fix an issue with k8s-events posted for log messages despite the configured levels.

> Issue : fixes #188

## Description

Configured log levels was ignored in the connector of the `logging` facilities to the k8s-event posting engine, so all events were posted. The docs were written for the event-posting shortcuts, but this was not taken into account when the log-posting engine was added.

In addition, some extra checks are added to verify that the shortcuts actually obey the configured levels.

**Note** that the debug level messages are now filtered not by the K8s-posting handler `K8sPoster` — the minimum level is removed from there— but by its `filter()` method, based on the Kopf's configs. Which is generally equivalent, as it was previously filtered by level in `super().filter(record)`, which stands in the same line.

## Types of Changes

- Bug fix (non-breaking change which fixes an issue)

## Review
_List of tasks the reviewer must do to review the PR_
- [ ] Tests
- [ ] Documentation

Issue #23, opened 2019-04-06
Configurable field for the status storage


nolar:
## Actual Behavior

Currently, Kopf stores the internal status of the handlers in `status.kopf` (hard-coded). It is used for exchanging the information cross the event cycles.

This is done in the assumption that there is only one operator/controller per resource kind.

If two or more different Kopf-based operators/controllers are handling the resource, especially the reusable resource, such as Pods, they can collide.

## Expected Behavior

The field of the internal status must be configurable. For example, in the handler declaration:

```python
@kopf.on.update('', 'v1', 'pods', status='status.kopf.some-other-field')
def pod_updated(**_):
    pass
```

The controlling parts of the custom resource A can have a convention to use its own resource kinds as the field:

```python
@kopf.on.update('', 'v1', 'pods', status='status.kopf-for-resource-a')
def pod_updated(**_):
    pass
```

If explicitly set to ``None``, the status is not persisted. Which implies that a different flow should be used (all-at-once lifecycle, errors ignored):

```python
@kopf.on.create('', 'v1', 'pods', status=None)
@kopf.on.update('', 'v1', 'pods', status=None)
@kopf.on.delete('', 'v1', 'pods', status=None)
def pod_event(**_):
    pass
```

---

PS: There is also the `metadata.annotations.last-seen-state`. It should be turned off when the status is turned off. It makes no sense to store the last-seen-state and to calculate the diff, since with the status not persisted, there will be no multiple handler calls.

nolar:
Solved alternatively with #331 and is available since `kopf>=0.27`

Docs: 

* https://kopf.readthedocs.io/en/latest/configuration/#handling-progress
* https://kopf.readthedocs.io/en/latest/configuration/#change-detection
* https://kopf.readthedocs.io/en/latest/configuration/#storage-transition

PR #336, opened 2020-03-26
Settings (official and documented)

nolar:
## What do these changes do?

Introduce settings: an official and documented way of configuring of all the little aspects of the framework behaviour on a per-operator basis.

## Description

Previously, module-level hacks were used to modify global variables grouped into classes-as-namespaces (beside being undocumented):

```python
import logging
import kopf

# DEPRECATED:
kopf.config.EventsConfig.events_loglevel = logging.WARNING
kopf.config.WatchersConfig.default_stream_timeout = 10 * 60
kopf.config.WorkersConfig.synchronous_tasks_threadpool_limit = 100

@kopf.on.create(...)
def create_fn(**_):
    pass
```

No more!

Settings are now a part of the framework, and should be defined in the startup handlers:

```python
import logging
import kopf

@kopf.on.startup()
def configure(settings, **_):
    settings.posting.level = logging.WARNING
    settings.watching.stream_timeout = 10 * 60
    settings.execution.limit = 100

@kopf.on.create(...)
def create_fn(**_):
    pass
```

And there are many more settings (migrated from the existing config classes), and all the new settings will be added to this structure (e.g. a state persistence storage from #331).

```python
import kopf

# NOT YET! SOON. See #331. 
@kopf.on.startup()
def configure(config: kopf.OperatorConfig, **_):
    config.persistence.storage = kopf.AnnotationsStateStorage(prefix='my-opr.example.com')
    # OR:
    config.persistence.storage = kopf.StatusStateStorage(name='my-opr')
```

## Backward-compatibility

The legacy way via module-level config-classes is supported — they are used as the defaults for the settings. BUT: runtime changes of the config-classes are ignored, only the runtime changes of the settings are obeyed.

A special case with the executor of synchronous handlers is supported, and calls to `set_synchronous_tasks_threadpool_limit()` are forwarded to the settings executor — until config-classes are removed with the next major release (1.0).

--- 

Beside exposing settings to the users officially, this change also adds a dependency injection mechanics which can be used in the tests instead of mocks — thus making the tests cleaner.

## Issues/PRs

> Issues: #23 

> Related: #322 #192 #109 

## Type of changes

## Checklist


docs/configuration.rst
```diff
@@ -0,0 +1,135 @@
...
+=============
+
+There are tools to configure some of kopf functionality, like asynchronous
+tasks behaviour and logging events.
+
+
+Startup configuration
+=====================
+
+Every operator has its settings (even if there are more than one operator
```
michaelnarodovitch: ... even if there is more then one...

nolar:
@haikoschol Thank you. I've implemented these recommendation with few little additions in #337.

Issue #321, opened 2020-03-03
Store handler progress data in an annotation, not the status subresource


lack:
## Problem

As of ``apiextensions.k8s.io/v1`` it is much more difficult to create a CRD that allows unvalidated information, as an openapi schema is required for even the status subresource.  Tracking the handler progress in an annotation instead of the status subresource means it's easier to use kopf to write operators for CRDs that we don't own.

To define a wide-open status subresource in a v1 CRD, you can do the following:

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: kopfexamples.zalando.org
spec:
  scope: Namespaced
  group: zalando.org
  versions:
    - name: v1
      served: true
      storage: true
      subresources:
        status: {}
      schema:
        openAPIV3Schema:
          type: object
          properties:
            spec:
              type: object
              properties:
                # Main spec validation schema...
            status:
              type: object
              x-kubernetes-preserve-unknown-fields: true
```

Or for something more structured you can allow only the kopf property of the status subresource to be unvalidated as follows:

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: kopfexamples.zalando.org
spec:
  scope: Namespaced
  group: zalando.org
  versions:
    - name: v1
      served: true
      storage: true
      subresources:
        status: {}
      schema:
        openAPIV3Schema:
          type: object
          properties:
            spec:
              type: object
              properties:
                # Main spec validation schema...
            status:
              type: object
              properties:
                result:
                  type: string
                  description: The result from the handler
                  enum:
                  - created
                  - updated
                # Other structured subresource schema data...
                kopf:
                  type: object
                  x-kubernetes-preserve-unknown-fields: true
```

## Proposal

Kopf already stores important metadata in annotations including ``kopf.zalando.org/last-handled-configuration`` and ``kubectl.kubernetes.io/last-applied-configuration``.  Setting up the state.py mechanism to similarly use an annotation instead of the status subresource for its progress data should, I hope, be relatively straight-forward.

For resources whose CRDs can be altered, the other mechanisms to allow status.kopf.progress is an okay work-around.  But the v1 schema enforcement means that kopf can't store its progress data there any more if the CRD cannot be altered.

## Checklist

- [x] Many users can benefit from this feature, it is not a one-time case
- [x] The proposal is related to the K8s operator framework, not to the K8s client libraries

[...]

Issue #308, opened 2020-02-10
Retry is not happening after exception


gtsystem:
## Long story short

If an exception occur inside a handler for `volumeattachments` objects, the handler will never get retried.

Looking at the logs, it seems that kopf try to patch the object status with `{"status":{"kopf": {"processing": ...}}}`
This patch call succeed but the object status is not changed. I guess this is the reason no further attempts are made.
In theory the `/status` endpoint should be used to patch the status, but from my testing this have no effects as well. It seems like custom fields are not accepted inside `"status"`.

Possible fix: use an annotation to store kopf status?

## Description

<details><summary>The code snippet to reproduce the issue</summary>

test_va.py
```python
class MyException(Exception):
    pass

@kopf.on.create('storage.k8s.io', 'v1', 'volumeattachments', retries=3, backoff=1.0)
def eventual_failure_with_tracebacks(**kwargs):
    raise Exception("An error that is supposed to be recoverable.")
```

va.yaml
```
apiVersion: storage.k8s.io/v1
kind: VolumeAttachment
metadata:
  name: test-vaa
spec:
  attacher: xxx 
  nodeName: kind-worker
  source:
    persistentVolumeName: a-pv
status:
  attached: false
```

</details>

<details><summary>The exact command to reproduce the issue</summary>

```bash
kopf run test_va.py &
kubectl create -f va.yaml
```
</details>

<details><summary>The full output of the command that failed</summary>

```
[2020-02-10 10:29:19,209] kopf.reactor.activit [INFO    ] Initial authentication has been initiated.
[2020-02-10 10:29:19,232] kopf.activities.auth [INFO    ] Handler 'login_via_pykube' succeeded.
[2020-02-10 10:29:19,245] kopf.activities.auth [INFO    ] Handler 'login_via_client' succeeded.
[2020-02-10 10:29:19,245] kopf.reactor.activit [INFO    ] Initial authentication has finished.
[2020-02-10 10:29:19,276] kopf.engines.peering [WARNING ] Default peering object not found, falling back to the standalone mode.
[2020-02-10 10:29:45,516] kopf.objects         [ERROR   ] [None/test-vaa] Handler 'eventual_failure_with_tracebacks' failed with an exception. Will retry.
Traceback (most recent call last):
  File "/src/_envl/lib64/python3.7/site-packages/kopf/reactor/handling.py", line 529, in _execute_handler
    lifecycle=lifecycle,  # just a default for the sub-handlers, not used directly.
[...]

PR #339, opened 2020-04-01
Add a warning if patch doesn't succeed

michaelnarodovitch:
## What do these changes do?

The situation
- kopf submits a `merge-patch+json` patch
- kubernetes API doesn't apply the `merge-patch+json` patch but responds `200 OK`

leads to bugs and confusion, when testing operators.

With this change, kopf will check, if the `merge-patch+json` patch was merged and be verbose, if the patch was not exactly successful.

## Description

**Before:** When a patch wasn't merged, then kopf didn't report anything.

**After:** When a patch wasn't merged, then kopf will report if ran in `--verbose` mode.

This will help users to debug their operators. Especially, when they use OpenAPI schemas in their CRDs.

Not sure if the check is in the right place and the logger is setup appropriately.

No risky changes.

## Issues/PRs

 

If a patch is not merged, then it is usually because of a schema mismatch of the CRD. For this reason it was discussed to move the internal kopf `status` to the annotations.  #308 #321

## Type of changes

## Checklist

### DIFF
--- a/kopf/clients/patching.py
+++ b/kopf/clients/patching.py
@@ -1,12 +1,17 @@
-from typing import Optional, cast
+import copy
+from typing import Optional, cast, Any, Union
 
 import aiohttp
 
 from kopf.clients import auth
 from kopf.clients import discovery
-from kopf.structs import bodies
+from kopf.structs import bodies, diffs
 from kopf.structs import patches
 from kopf.structs import resources
+
+from logging import getLogger
+
+logger = getLogger('kopf.clients.patching')
 
 
 @auth.reauthenticated_request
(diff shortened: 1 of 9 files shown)


kopf/clients/patching.py
```diff
@@ -1,13 +1,18 @@
-from typing import Optional, cast
+import copy
+from typing import Optional, cast, Any, Union
 
 import aiohttp
 
 from kopf.clients import auth
 from kopf.clients import discovery
-from kopf.structs import bodies
+from kopf.structs import bodies, diffs
```
nolar, line 8: It is better to keep the code style consistent with the surroundings. I regularly (once a week or two) think on switching to this PyCharm-suggested form of one-line imports (as it does on every refactoring or move moves), but this should be done as a separate PR for all the modules — for consistency.

Meanwhile, let's follow the existing [Google Style Guide for imports](http://google.github.io/styleguide/pyguide.html#313-imports-formatting).

michaelnarodovitch:
Thx for all the comments and suggestions. Had to add quite a bit of code. I think quality is OK and changes are safe.

Missing is integration tests with the Kubernetes API responses, and especially integration tests with the `status_patch` merging. What I did was based of an educated guess from reviewing https://github.com/kubernetes/apiextensions-apiserver. If we wanted to be 100% safe, we would also have to test the `PATCH` API of the canonical Kubernetes objects.

kopf/clients/patching.py
michaelnarodovitch: Actually, I like the google style more as it produces cleaner diffs (one line per unit change). Can be checked automatically with https://github.com/PyCQA/flake8-import-order . Pycharm can be configured to do google style imports https://www.jetbrains.com/help/pycharm/configuring-code-style.html.

nolar:
Implemented in https://github.com/nolar/kopf/pull/527

PR #346, opened 2020-04-09
Limit persistence annotations to 63 chars

nolar:
## Description

An issue found the hard way (though we hit it before): when an annotation's name is 64 or more characters, it is not accepted by K8s API. This is [documented](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set).

For example if we have a prefix, a long function name, and maybe a field suffix: `kopf.zalando.org/some_resource_of_interest_created.status.childname` — and oops, it is 67 already, while 63 is allowed.

This PR fixes the issue with lengthy handler ids by cutting their tail, and replacing it with some consistent one-way hash — this makes the annotation unique enough, and still recognisable/readable for the primary part of its name.

## Issues/PRs

> Issues: #331 

## Type of changes

## Checklist

### DIFF
--- a/kopf/storage/progress.py
+++ b/kopf/storage/progress.py
@@ -39,10 +39,11 @@ with the following keys:
 All timestamps are strings in ISO8601 format in UTC (no explicit ``Z`` suffix).
 """
 import abc
+import base64
 import copy
-import datetime
+import hashlib
 import json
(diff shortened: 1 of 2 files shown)


stehem:
:+1:

PR #351, opened 2020-04-27
Throttle individual resource processing on unexpected errors

nolar:
## What do these changes do?

Whenever an unexpected error happens for an individual resource (in the framework/environment/API, not in the handlers), then pause for an increasing delay — until self-recovered.

## Description

This issue was detected in an incident when PATCH request failed due to HTTP 422 "Unprocessable Entity" (#346). Instead of stopping or slowing down any attempts, the operator continued handling repeatedly with 1-2 attempts per second.

On a wider scope, if _anything_ goes wrong for an individual resource, slow down processing of this individual resource to reduce the damage.

The slow down is incrementing: from 1 second to 10 minutes, and then remains at 10 mins until recovered.

This throttling also cover the resource PATCH'ing if something is wrong with the API itself, or with the content of the patches.

All other individual resources of all kinds continue their processing as usual.

The handlers' errors are handled in their own way with backoff intervals and retry limits/timeouts — this is not changed. This fix is about the outside-of-the-handlers codebase and environment issues.

The safety net covers the codebase from the resource processing begins (slightly after this: when the per-resource memory is restored/created), and until the handlers. The codebase from watchers to the processing is covered by #350 — the operator exits in that case (global or per-kind, but not per-resource errors).

## Issues/PRs

> Issues: #346 

> Related: #331 #350 

## Type of changes

## Checklist

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -364,3 +364,36 @@ e.g. by arbitrarily labelling them, so that a new diff-base is generated:
     kubectl label kex -l somelabel=somevalue  ping=pong
 
 Then, switch to the new storage alone, without the transitional setup.
+
+
+Error throttling
+================
+
+To prevent uncontrollable flood of activities in case of errors that prevent
+the resources being marked as handled, which could lead to Kubernetes API
+flooding, it is possible to throttle the activities on a per-resource basis:
+
+.. code-block:: python
+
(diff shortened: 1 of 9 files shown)


nolar:
Closed in favor of nolar/kopf#510

PR #371, opened 2020-06-08
Deprecate old K8s versions (1.12)

nolar:
## What do these changes do?

Drop old K8s versions (up to 1.12) from testing.

## Description

Minikube does not support K8s 1.12 anymore, the builds fail: 

* https://travis-ci.org/github/zalando-incubator/kopf/jobs/696167018

This does not break the code itself, and the existing releases are fine, but all new PRs will fail at this test step.

So, we can drop the old enough version of K8s. Instead, enlist 1.17 explicitly, since the "latest" now refers to 1.18.

## Issues/PRs

> Issues: #13 

## Type of changes

- Mostly CI/CD automation, contribution experience

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,8 @@ env:
   matrix:
     - KUBERNETES_VERSION=latest CLIENT=yes  # only one "yes" is enough
     - KUBERNETES_VERSION=latest CLIENT=no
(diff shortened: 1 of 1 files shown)


nolar:
Closed in favor of https://github.com/nolar/kopf/pull/501

Issue #391, opened 2020-08-14
client_timeout on watch results in a fatal error


jscaltreto:
## Long story short

If settings.client_timeout is set and the timeout is reached while watching, aiohttp.TimeoutError is raised and not caught resulting in the operator dying. I believe that timeouts should be treated like other connection errors hit while watching; that is, they are kept silent and the watch is allowed to restart on its own.

## Traceback
```
Traceback (most recent call last):
  File "/Users/jscaltreto/git/operator/build/.venv/lib/python3.7/site-packages/kopf/clients/watching.py", line 226, in watch_objs
    async for line in _iter_jsonlines(response.content):
  File "/Users/jscaltreto/git/operator/build/.venv/lib/python3.7/site-packages/kopf/clients/watching.py", line 269, in _iter_jsonlines
    async for data in content.iter_chunked(chunk_size):
  File "/Users/jscaltreto/git/operator/build/.venv/lib/python3.7/site-packages/aiohttp/streams.py", line 39, in __anext__
    rv = await self.read_func()
  File "/Users/jscaltreto/git/operator/build/.venv/lib/python3.7/site-packages/aiohttp/streams.py", line 368, in read
    await self._wait('read')
  File "/Users/jscaltreto/git/operator/build/.venv/lib/python3.7/site-packages/aiohttp/streams.py", line 296, in _wait
    await waiter
  File "/Users/jscaltreto/git/operator/build/.venv/lib/python3.7/site-packages/aiohttp/helpers.py", line 596, in __exit__
    raise asyncio.TimeoutError from None
concurrent.futures._base.TimeoutError
```

jscaltreto:
moving to nolar/kopf

PR #392, opened 2020-08-14
Treat client timeouts during watches similarly to other http errors

jscaltreto:
## What do these changes do?

`asyncio.TimeoutError` is caught and ignored like other http errors

## Description

Currently, if `connect_timeout` or `client_timeout` is reached, the error is not caught and is effectively fatal; the operator needs to be restarted. However, some other http errors are ignored and the watch is resumed by `infinite_watch()`. I believe this may have been an oversight as I can see no reason why a timeout should be considered a fatal exception when other http errors are not.

## Issues/PRs
#391 

## Type of changes

## Checklist

   - There were no existing unit tests for the silent error handling as noted in the comment to infinite_watch()


jscaltreto:
moving to nolar/kopf

